PR #2, opened 2016-07-18
(TK-384) Initial code

KevinCorcoran:
This pull request contains my code from https://github.com/puppetlabs/pe-file-sync/pull/257, @cprice404's code from https://github.com/cprice404/pe-file-sync/commit/a926e3ce15a4612b0245d1c0ed3ed4dce160e19f#diff-9e25f3a6c56b4a21bd72fde5efd54313, and I think I've addressed all of the great feedback you guys gave me on the previous PR.

The only thing in here that I think should be of any surprise is 79d287147357223c03be171395f98522606837ec.  I included that because it turned out to be pretty easy to do; I'm reasonably happy with how it turned out, but it's also pretty different than any other TK services I've seen before, so I'm not dead-set on that approach.  I wouldn't mind just ripping it out if y'all don't like it - shouldn't be too hard to burn down.

Other than that, I've filed a separate ticket about the whole mess of `.poll` vs. `.take`, `future` vs. scheduler and whatnot.

I think this probably deserves at least a little more test coverage than what it has right now, especially a test case for multiple watchers, which isn't covered at all right now.  But, otherwise, I'm hoping this is close-ish to merge-able.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -22,13 +22,23 @@
 
   :plugins [[lein-release "1.0.5"]]
 
-  :lein-release {:scm         :git
-                 :deploy-via  :lein-deploy}
+  :lein-release {:scm :git
+                 :deploy-via :lein-deploy}
 
   :deploy-repositories [["releases" {:url "https://clojars.org/repo"
                                      :username :env/clojars_jenkins_username
                                      :password :env/clojars_jenkins_password
                                      :sign-releases false}]]
+
+  :source-paths ["src/clj"]
+  :java-source-paths ["src/java"]
+
+  :profiles {:dev {:dependencies [[puppetlabs/trapperkeeper ~tk-version
+                                   :classifier "test"
+                                   :scope "test"]
+                                  [puppetlabs/kitchensink ~ks-version
+                                   :classifier "test"
+                                   :scope "test"]]}}
 
   :main puppetlabs.trapperkeeper.main
 )
--- a/project.clj
+++ b/project.clj
@@ -20,11 +20,6 @@
                  [puppetlabs/trapperkeeper-scheduler "0.0.1"]
                  [puppetlabs/i18n "0.4.1"]]
 
-  :plugins [[lein-release "1.0.5"]]
-
-  :lein-release {:scm :git
-                 :deploy-via :lein-deploy}
-
   :deploy-repositories [["releases" {:url "https://clojars.org/repo"
                                      :username :env/clojars_jenkins_username
                                      :password :env/clojars_jenkins_password
--- a/src/clj/puppetlabs/services/protocols/filesystem_watch_service.clj
+++ b/src/clj/puppetlabs/services/protocols/filesystem_watch_service.clj
@@ -12,7 +12,11 @@
     "Given a directory on the filesystem, initiate watching of dir.  The
     watcher's callbacks will be invoked when dir changes.  Available options are:
       * :recursive true   - If true, callbacks will be invoked when dir or any
-                            file underneath dir changes.")
+                            file underneath dir changes.
+
+    If dir is deleted, the watch is invalidated; if dir is recreated, the
+    watcher's callbacks will not be invoked unless add-watch-dir! is called
+    a second time to re-establish the watch on dir.")
 
   (add-callback! [this callback]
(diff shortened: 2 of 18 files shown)


project.clj
```diff
@@ -22,13 +22,23 @@
 
   :plugins [[lein-release "1.0.5"]]
 
-  :lein-release {:scm         :git
-                 :deploy-via  :lein-deploy}
+  :lein-release {:scm :git
+                 :deploy-via :lein-deploy}
```
cprice404: I think we actually don't need this boilerplate anymore; it was for the old lein-release plugin, and {{release}} is a task built in to lein itself now.  I think you can also get rid of the reference to the plugin above (and in any other repos you see this stuff still hanging around in).

cprice404:
Travis failures seem like they might be a real thing, guessing another difference between osx-- and linux.

cprice404:
+1 other than travis failures, your API change seems ~legit to me

cprice404:
oh, and yeah, a test or two with multiple watchers would be good, but I wouldn't necessarily block the PR on that if it's holding up other tickets.

KevinCorcoran:
@cprice404 Thanks for the review, I pushed 2 commits which I think address your feedback.  I'm going to keep working on some more test coverage but you I agree with your point that this could probably be merged without the extra tests in the interest of un-blocking other work.

KevinCorcoran:
I added one more commit with the extra test coverage.  I'm hoping you and @camlow325 find it to be sufficient ... this stuff seems like I could write tests for it until the cows come home and I'd still be able to come up with more tests to write.

src/clj/puppetlabs/services/watcher/filesystem_watch_service.clj
```diff
@@ -0,0 +1,45 @@
+(ns puppetlabs.services.watcher.filesystem-watch-service
```
haus: This looks inconsistent with other tk services. looks like they mostly start with
`puppetlabs.trapperkeeper.services.` and not `puppetlabs.services.`

src/clj/puppetlabs/services/protocols/filesystem_watch_service.clj
```diff
@@ -0,0 +1,38 @@
+(ns puppetlabs.services.protocols.filesystem-watch-service
```
haus: Same as a comment in the implementation, but other tk service protocols use the `puppetlabs.trapperkeeper.services.protocols.` namespace. Seems like consistency would be good there.

test/puppetlabs/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -0,0 +1,387 @@
...
+         (testing "Initial directory creation and deletion"
+           (is (fs/mkdir sub-dir))
+           (let [events #{{:path sub-dir
+                           :type :create}}]
+             (is (= events (wait-for-events results events))))
+           (is (fs/delete sub-dir))
+           (let [events #{{:path sub-dir
+                           :type :delete}}]
+             (is (= events (wait-for-events results events)))))
+         (testing "Re-creating the directory fires an event as expcted"
```
camlow325: s/expcted/expected

camlow325:
👍 from me.  I didn't have a chance to review all of the tests yet but what I did looked good to me.  I'm good with this being merged when @haus feedback has been addressed.

src/clj/puppetlabs/services/watcher/filesystem_watch_service.clj
KevinCorcoran: Oh.  Interesting.  To start, I followed the pattern in file sync, and then remove an `enterprise` level when I moved the code into this repo.

Kinda weird, though, to have things in `puppetlabs.trapperkeeper` which aren't part of Trapperkeeper itself.  But I guess that ship has sailed.

KevinCorcoran:
@cprice404 @camlow325 @haus I think we're pretty much at consensus on this (I recently added 2 more commits to address the last bits of feedback).  I'm going to self-merge this so I can do an initial release to Clojars, but I'm happy to address any other issues you should find in a subsequent PR.

PR #4, opened 2016-07-27
(TK-385) Use .take instead of .poll for Watchers

justinstoller:
Prior to this we polled each Watcher registered every second for
updated WatchKeys.

This patch moves us to use a blocking .take on each Watcher in a
backgrounded future.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -53,7 +53,9 @@
 (defn validate-watch-options!
   [options]
   (when-not (= true (:recursive options))
-    (throw (IllegalArgumentException.(trs "Support for non-recursive directory watching not yet implemented")))))
+    (throw
+      (IllegalArgumentException.
+        (trs "Support for non-recursive directory watching not yet implemented")))))
 
 (defrecord WatcherImpl
   [watch-service callbacks]
@@ -69,8 +71,9 @@
 
 (defn create-watcher
   []
-  (map->WatcherImpl {:watch-service (.newWatchService (FileSystems/getDefault))
-                     :callbacks (atom [])}))
+  (map->WatcherImpl
+    {:watch-service (.newWatchService (FileSystems/getDefault))
+     :callbacks (atom [])}))
 
 (schema/defn watch-new-directories!
   [events :- [Event]
@@ -102,7 +105,8 @@
               (log/debug (trs "Events:\n{0}"
                               (pprint-events events)))
               (log/trace (trs "orig-events:\n{0}"
-                              (ks/pprint-to-string (map clojurize-for-logging orig-events))))
+                              (ks/pprint-to-string
+                                (map clojurize-for-logging orig-events))))
               (shutdown-on-error #(doseq [callback callbacks]
                                     (callback events)))
               (watch-new-directories! events watcher)
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -21,6 +21,10 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Functions
+;;;
+;;; Helper functions in this namespace are heavily influenced by WatchDir.java
+;;; from Java's official documentation:
+;;; https://docs.oracle.com/javase/tutorial/essential/io/notification.html
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (schema/defn clojurize :- Event
@@ -88,29 +92,22 @@
    shutdown-on-error :- IFn]
   (future
     (while (not @stopped?)
-      ;; `.take` here will block until there has been an event for this key
-      ;;
-      ;; If `.close` is called on the `WatchService` while blocking on
-      ;; `.take` a `ClosedWatchServiceException` will be raised.
-      ;; The `WatchKey` will then be invalidated, so `.reset` will return
-      ;; false, but enqueued events will still be available via `.pollEvents`
-      ;; and `.watchable` should continue to function as desired.
-      (let [watch-key (try
-                        (.take (:watch-service watcher))
-                        (catch ClosedWatchServiceException e
-                          (log/info (trs "Closing watcher {0}" watcher))))
-            orig-events (.pollEvents watch-key)
-            events (map #(clojurize % (.watchable watch-key)) orig-events)
-            callbacks @(:callbacks watcher)]
-        (when-not (empty? events)
-          (log/info (trs "Got {0} event(s) for watched-path {1}"
-                         (count orig-events) (.watchable watch-key)))
-          (log/debug (trs "Events:\n{0}"
-                          (pprint-events events)))
-          (log/trace (trs "orig-events:\n{0}"
-                          (ks/pprint-to-string (map clojurize-for-logging orig-events))))
-          (shutdown-on-error #(doseq [callback callbacks]
-                               (callback events)))
-          (watch-new-directories! events watcher)
-          (.reset watch-key))))))
+      (try
+        (let [watch-key (.take (:watch-service watcher))
+              orig-events (.pollEvents watch-key)
+              events (map #(clojurize % (.watchable watch-key)) orig-events)
+              callbacks @(:callbacks watcher)]
+          (when-not (empty? events)
+            (log/info (trs "Got {0} event(s) for watched-path {1}"
+                           (count orig-events) (.watchable watch-key)))
+            (log/debug (trs "Events:\n{0}"
+                            (pprint-events events)))
+            (log/trace (trs "orig-events:\n{0}"
+                            (ks/pprint-to-string (map clojurize-for-logging orig-events))))
+            (shutdown-on-error #(doseq [callback callbacks]
+                                  (callback events)))
+            (watch-new-directories! events watcher)
+            (.reset watch-key)))
+       (catch ClosedWatchServiceException e
+         (log/info (trs "Closing watcher {0}" watcher)))))))
 
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -102,11 +102,13 @@
             (when-not (empty? events)
               (log/info (trs "Got {0} event(s) for watched-path {1}"
                              (count orig-events) (.watchable watch-key)))
-              (log/debug (trs "Events:\n{0}"
-                              (pprint-events events)))
-              (log/trace (trs "orig-events:\n{0}"
-                              (ks/pprint-to-string
-                                (map clojurize-for-logging orig-events))))
+              (log/debug (format "%s\n%s"
+                                 (trs "Events:")
+                                 (pprint-events events)))
+              (log/trace (format "%s\n%s"
+                                 (trs "orig-events:")
+                                 (ks/pprint-to-string
+                                   (map clojurize-for-logging orig-events))))
               (shutdown-on-error #(doseq [callback callbacks]
                                     (callback events)))
               (watch-new-directories! events watcher)
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -84,30 +84,30 @@
                                           (map #(.toPath (:path %)))))))
 
 (schema/defn watch!
-  "Returns a future representing processing events for the passed in watcher.
-  The future will swallow errors, loop producing side-effects until stopped,
-  and is expected to coordinate with other functions via the stopped? atom."
+  "Creates a future and processes events for the passed in watcher.
+  The future will continue until the underlying WatchService is closed."
   [watcher :- (schema/protocol Watcher)
-   stopped? :- Atom
    shutdown-on-error :- IFn]
   (future
-    (while (not @stopped?)
-      (try
-        (let [watch-key (.take (:watch-service watcher))
-              orig-events (.pollEvents watch-key)
-              events (map #(clojurize % (.watchable watch-key)) orig-events)
-              callbacks @(:callbacks watcher)]
-          (when-not (empty? events)
-            (log/info (trs "Got {0} event(s) for watched-path {1}"
-                           (count orig-events) (.watchable watch-key)))
-            (log/debug (trs "Events:\n{0}"
-                            (pprint-events events)))
-            (log/trace (trs "orig-events:\n{0}"
-                            (ks/pprint-to-string (map clojurize-for-logging orig-events))))
-            (shutdown-on-error #(doseq [callback callbacks]
-                                  (callback events)))
-            (watch-new-directories! events watcher)
-            (.reset watch-key)))
-       (catch ClosedWatchServiceException e
-         (log/info (trs "Closing watcher {0}" watcher)))))))
+    (let [stopped? (atom false)]
+      (while (not @stopped?)
+        (try
+          (let [watch-key (.take (:watch-service watcher))
+                orig-events (.pollEvents watch-key)
+                events (map #(clojurize % (.watchable watch-key)) orig-events)
+                callbacks @(:callbacks watcher)]
+            (when-not (empty? events)
+              (log/info (trs "Got {0} event(s) for watched-path {1}"
+                             (count orig-events) (.watchable watch-key)))
+              (log/debug (trs "Events:\n{0}"
+                              (pprint-events events)))
+              (log/trace (trs "orig-events:\n{0}"
+                              (ks/pprint-to-string (map clojurize-for-logging orig-events))))
+              (shutdown-on-error #(doseq [callback callbacks]
+                                    (callback events)))
+              (watch-new-directories! events watcher)
+              (.reset watch-key)))
+         (catch ClosedWatchServiceException e
+           (reset! stopped? true)
+           (log/info (trs "Closing watcher {0}" watcher))))))))
 
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -102,13 +102,13 @@
             (when-not (empty? events)
               (log/info (trs "Got {0} event(s) for watched-path {1}"
                              (count orig-events) (.watchable watch-key)))
-              (log/debug (format "%s\n%s"
-                                 (trs "Events:")
-                                 (pprint-events events)))
-              (log/trace (format "%s\n%s"
-                                 (trs "orig-events:")
-                                 (ks/pprint-to-string
-                                   (map clojurize-for-logging orig-events))))
+              (log/debugf "%s\n%s"
+                          (trs "Events:")
+                          (pprint-events events))
+              (log/tracef "%s\n%s"
+                          (trs "orig-events:")
+                          (ks/pprint-to-string
+                            (map clojurize-for-logging orig-events)))
               (shutdown-on-error #(doseq [callback callbacks]
                                     (callback events)))
               (watch-new-directories! events watcher)
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -6,15 +6,12 @@
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.trapperkeeper.services.protocols.filesystem-watch-service :refer [Event Watcher]])
   (:import (clojure.lang Atom IFn)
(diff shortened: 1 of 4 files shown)


KevinCorcoran:
Travis CI failed and looks like legitimate problem with the code and/or tests.

When I did the initial work on this project I found quite a few differences between the OSX and Linux implementations of `WatchService` and therefore found it sometimes necessary to keep a Linux VM handy to test the code on as I was developing it.  https://github.com/puppetlabs/trapperkeeper-filesystem-watcher#development

I ended up just using rsync to sync changes over and then run the tests, example command:

```
$ rsync -avz project.clj dev dev-resources src test  p5jkbt64pq06qar.delivery.puppetlabs.net:pe-file-sync
```

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,51 +79,29 @@
...
                                           (map #(.toPath (:path %)))))))
 
-(defn reset-key-and-unwatch-deleted!
-  [watch-key]
-  ;; reset key and remove from set if directory no longer accessible
-  (let [valid? (.reset watch-key)]
-    (when (not valid?)
-      ;; TODO: not sure if we need to do anything else here now that we're not retaining a map of
-      ;; these keys.  The Oracle example code makes it seem like maybe you do, but I can't tell what.
-      (log/info (trs "Removing watched directory {0} because it was deleted" (.watchable watch-key))))))
```
KevinCorcoran, line 92: Why was this deleted?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,51 +79,29 @@
...
-
-;; This function (and other helper functions in this namespace)
-;; are heavily influenced by WatchDir.java from Java's official documentation:
-;; https://docs.oracle.com/javase/tutorial/essential/io/notification.html
```
KevinCorcoran, line 96: This comment still seems valid and I think it would be nice to keep it around ... any particular reason you deleted it?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,51 +79,29 @@
...
-  ;; nested under the root watched path) but the extent to which that is
-  ;; true is likely platform-dependent.
-  ;;
-  ;; https://tickets.puppetlabs.com/browse/PE-15621
-
-  (doseq [watcher watchers]
-    (when-let [watch-key (.poll (:watch-service watcher))]
-      (let [orig-events (.pollEvents watch-key)
+  (future
+    (while (not @stopped?)
```
KevinCorcoran: It looks like `stopped?` is currently only read and never written to.  I think you need to add something like `(reset! stopped? true)` inside the `catch` expression a few lines below.

Also, it looks like `stopped?` could be a local binding inside this function (in a `let` expression immediately above) instead of part of the TK service context.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran, line 92: Oh, I see, you just moved the call to `.reset` into `watch!`

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,51 +79,29 @@
...
+      ;; `.take` here will block until there has been an event for this key
+      ;;
+      ;; If `.close` is called on the `WatchService` while blocking on
+      ;; `.take` a `ClosedWatchServiceException` will be raised.
+      ;; The `WatchKey` will then be invalidated, so `.reset` will return
+      ;; false, but enqueued events will still be available via `.pollEvents`
+      ;; and `.watchable` should continue to function as desired.
```
KevinCorcoran: Your comment makes sense, but I think it would be better to not execute the rest of this loop if this exception is caught.

If this exception is caught, that means that the service's `stop` function executed and the application is shutting down.  It's possible that the other services which registered callbacks have already shutdown themselves, in which case, invoking their event handler callbacks might not even be a safe thing to do.

The code is certainly a little cleaner like this, but I don't think it's the correct behavior.  It might be sufficient to simply move the rest of the body of this function inside `try`.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: `stopped?` is set by the `stop` [service function](https://github.com/justinstoller/trapperkeeper-filesystem-watcher/blob/c8d7b6070599d1e85320437338a8c941d30d7629/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service.clj#L25) before calling `.close` on the watcher.

In theory, by the time we get to the `catch` block `stopped?` should already be `true`.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller, line 96: I felt like I was heavily changing the function that it was above and so it might no longer be applicable. In a new commit I've removed the bit about 'This function' and placed the note at the top of the functions section. The link to the oracle docs was very helpful!

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: That makes a lot of sense, updated in 0e00420

justinstoller:
We just got lucky with Travis passing. Nothing in the new commit should have affected the tests. The original failures in Travis, after much fighting with them, seem to all be existing race conditions in the tests and a [bug in the JVM](https://bugs.openjdk.java.net/browse/JDK-8145981) that are all much more prevalent when running on Linux now that we aren't polling.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran: I see.

Still, though, it seems like it'd be an improvement to move that inside the `future` in the interest of minimizing service state and keeping it local to that function.  Along with that, we could write a test that it behaves as expected (especially the part about `.take` throwing a `ClosedWatchServiceException`) - which would be a nice assurance to have.

KevinCorcoran:
I agree that this shouldn't affect the test failures, but, just to be sure: have you observed the tests failing on this commit?

KevinCorcoran:
I've seen the tests fail on this commit on both OSX and Linux.  The OSX failure is perhaps more interesting, since so far the intermittent failures have been isolated to Linux AFAIK.  Here's the output:

```
$ lein test

timed-out waiting for events

lein test :only puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service-test/multi-callbacks-test

FAIL in (multi-callbacks-test) (filesystem_watch_service_test.clj:167)
both callbacks invoked when both roots change
expected: (= events-2 (wait-for-events results-2 events-2))
  actual: (not (= #{{:path #object[java.io.File 0x5995ba6c "/var/folders/44/dd9x_8fj5xj90xs8z_s35dqc0000gq/T/multi-root-test-21469790799093-4094900126/test-file"], :type :modify}} []))

Ran 6 tests containing 52 assertions.
1 failures, 0 errors.
```

That is happening only intermittently, not every run.

Also - far less important, but worth mentioning - it appears that the error `java.lang.Exception: boom` is still being logged on Linux (but not OSX ... ugh).

KevinCorcoran:
I've been unable to reproduce that error on OSX in about 10 additional test runs.

My guess is that it occurred simply because the test lost the race, since the actual events which occurred was empty (`[]`).  This is nice because it seems unrelated to the JDK bug(s) we're fighting right now, but it probably means that we want to bump up `wait-time` at some point.

KevinCorcoran:
The hits keep on coming ...

On Linux, the test which was failing most often for me is `multiple-watch-dirs-test`.  However, when I run just that test, not the whole suite, I can't get it to fail ... even after about a dozen attempts.  Fun.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,80 +83,31 @@
...
-                                     after
-                                     stopped?
-                                     shutdown-on-error))))
-
-(schema/defn ^:always-validate schedule-watching!
-  [watchers :- (schema/atom [(schema/protocol Watcher)])
-   after :- IFn
+(schema/defn watch!
+  "Returns a future representing processing events for the passed in watcher.
+  The future will swallow errors, loop producing side-effects until stopped,
```
KevinCorcoran: Are the errors logged?  They should be.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran: After noticing that test case was failing only when run as part of the full suite, I got concerned that perhaps the existing shutdown code wasn't working as it should.  So, I implemented the change I was suggesting here: https://github.com/KevinCorcoran/trapperkeeper-filesystem-watcher/commit/e51d8050ff9937a2092fe2512230474bbbc4878f?w=1

After getting that done and looking at it, I'm pleased with how the production code change worked out.  The tests kinda suck because they just scape the log, but I still think they are worth while as a cursory level of assurance that our shutdown is operating as expected.  I think you should add that commit to this PR - assuming you agree.

Unfortunately, these changes don't seem to help the intermittent test failures at all.

KevinCorcoran: Ugh, also this: https://github.com/KevinCorcoran/trapperkeeper-filesystem-watcher/commit/6101a7b42dd8020824a8eae233648c44787ace10

That test is all race conditions in and log scraping and (to contradict my previous comment) I'm losing faith that it's actually worth having.

KevinCorcoran:
I tried increasing the `wait-time` to 30 seconds, but was still able to reproduce the `multi-callbacks-test` test failure mentioned above, in which the expected event just doesn't show up.

KevinCorcoran:
Even though some intermittent test failures have been observed on the code currently in `master`, it's clear that this PR greatly excacerbates the situation.  In local testing on Linux I usually get at least one failure on the code in this PR - whereas I was just able to do 8 consecutive runs on `master` without a failure.

KevinCorcoran:
Another observation along those lines: the test error `java.lang.Exception: boom` is not getting logged when running the tests on the code in `master`.  I'm kind of stumped what it is about these changes that's causing that to show up now.

KevinCorcoran:
Debugging the failures in `multiple-watch-dirs-test`, there seem to be distinct kinds: the first looks like this:

```
FAIL in (multiple-watch-dirs-test) (filesystem_watch_service_test.clj:301)
Watching of multiple directories using a single watcher Nested watch directories

expected: #{{:path #<java.io.File@353a829f /tmp/dir-11469803257499-3221284083/nested-dir>,
             :type :create}
            {:path #<java.io.File@320f9b40 /tmp/dir-11469803257499-3221284083/nested-dir/test-file>,
             :type :create}}
  actual: ({:path #<java.io.File@1e6e59d8 /tmp/dir-21469803257499-2439609755/test-file>,
            :type :modify}
           {:path #<java.io.File@21084cbe /tmp/dir-21469803257499-2439609755/nested-dir>,
            :type :create}
           {:path #<java.io.File@55204c19 /tmp/dir-11469803257499-3221284083/nested-dir/test-file>,
            :type :create}
           {:path #<java.io.File@7b3eea4e /tmp/dir-11469803257499-3221284083/nested-dir/test-file>,
            :type :modify})
```

The relevant part of that diff is this:

```
expected: #{{:path #<java.io.File@353a829f /tmp/dir-11469803257499-3221284083/nested-dir>,
             :type :create}
```

... vs ...

```
           {:path #<java.io.File@21084cbe /tmp/dir-21469803257499-2439609755/nested-dir>,
            :type :create}
```

The only relevant difference there is that the file is rooted `dir-2`  instead of `dir-2`.  This looks like the JDK bug.

The other kind of failure looks like this:

```
FAIL in (multiple-watch-dirs-test) (filesystem_watch_service_test.clj:301)
Watching of multiple directories using a single watcher Nested watch directories

expected: #{{:path #<java.io.File@3c02b068 /tmp/dir-11469803974337-3355536445/nested-dir>,
             :type :create}
            {:path #<java.io.File@a91212e /tmp/dir-11469803974337-3355536445/nested-dir/test-file>,
             :type :create}}
  actual: ({:path #<java.io.File@38ec1c34 /tmp/dir-21469803974337-1648469250/test-file>,
            :type :modify}
           {:path #<java.io.File@2a144ea0 /tmp/dir-11469803974337-3355536445/nested-dir>,
            :type :create}
           {:path #<java.io.File@33c2a7a /tmp/dir-11469803974337-3355536445/nested-dir/test-file>,
            :type :modify}
           {:path #<java.io.File@a7e362 /tmp/dir-11469803974337-3355536445/test-file>,
            :type :create})
```

In that case, the key difference seems to be that the test produced `:modify` event for `.../nested-dir/test-file` but not a `:create` event.  I don't know how to explain this.

KevinCorcoran:
Ugh, looks like that JDK bug can show up in `multi-callbacks-test` and cause it to fail in a new a different way:

```
2016-07-29 08:35:03,196 INFO  [p.t.s.w.filesystem-watch-core] Got 1 event(s) for watched-path /tmp/multi-root-test-11469806502847-3397460743
2016-07-29 08:35:03,201 DEBUG [p.t.s.w.filesystem-watch-core] Events:
({:type :modify,
  :path "/tmp/multi-root-test-11469806502847-3397460743/test-file"})

2016-07-29 08:35:03,210 TRACE [p.t.s.w.filesystem-watch-core] orig-events:
({:context #object[sun.nio.fs.UnixPath 0x7b3e3697 "test-file"],
  :count 1,
  :kind
  #object[java.nio.file.StandardWatchEventKinds$StdWatchEventKind 0x5ed4796f "ENTRY_MODIFY"]})

2016-07-29 08:35:03,211 DEBUG [p.t.s.w.filesystem-watch-core] Calling blocking .take on WatchService
2016-07-29 08:35:03,212 TRACE [p.t.s.w.filesystem-watch-core] orig-events:
({:context #object[sun.nio.fs.UnixPath 0x18143557 "test-file"],
  :count 1,
  :kind
  #object[java.nio.file.StandardWatchEventKinds$StdWatchEventKind 0x3ff264c "ENTRY_CREATE"]})


FAIL in (multi-callbacks-test) (filesystem_watch_service_test.clj:156)
callback-2 is invoked when root-2 changes but not callback-1
2016-07-29 08:35:03,215 DEBUG [p.t.s.w.filesystem-watch-core] Calling blocking .take on WatchService
2016-07-29 08:35:03,222 INFO  [p.t.s.w.filesystem-watch-core] Got 1 event(s) for watched-path /tmp/multi-root-test-21469806502848-2368811098
2016-07-29 08:35:03,227 DEBUG [p.t.s.w.filesystem-watch-core] Events:
({:type :modify,
  :path "/tmp/multi-root-test-21469806502848-2368811098/test-file"})


expected: ({:path #<java.io.File@3d69bbe3 /tmp/multi-root-test-11469806502847-3397460743/test-file>,
            :type :modify})
  actual: []
```

justinstoller:
One note @KevinCorcoran, you mention two failure modes from the `multiple-watch-dirs-test` where we get a `:modify` for the nested test file but a `:create` for a file in the parent directory. That output is a little misleading because the test case prior to it modifies a file by that name in the parent directory, so it can seems like that's the event we're getting (leaked from the previous test case). However, if we update the names of the nested file created we see that that `:create` is actually for the file in the nested dir, it's reported of having happened in the parent dir as a result, I think, again of the JDK bug:

```
FAIL in (multiple-watch-dirs-test) (filesystem_watch_service_test.clj:301)
Watching of multiple directories using a single watcher Nested watch directories

expected: #{{:path #<java.io.File@3e47d465 /tmp/dir-11469829400048-2161702959/nested-dir>,
             :type :create}
            {:path #<java.io.File@4d574ec7 /tmp/dir-11469829400048-2161702959/nested-dir/nested-test-file>,
             :type :create}}
  actual: ({:path #<java.io.File@2f9a624c /tmp/dir-11469829400048-2161702959/nested-dir>,
            :type :create}
           {:path #<java.io.File@39cdbc4e /tmp/dir-11469829400048-2161702959/nested-dir/nested-test-file>,
            :type :modify}
           {:path #<java.io.File@39a9b3ee /tmp/dir-11469829400048-2161702959/nested-test-file>,
            :type :create})
```

justinstoller:
> On Linux, the test which was failing most often for me is multiple-watch-dirs-test. However, when I run just that test, not the whole suite, I can't get it to fail ... even after about a dozen attempts. Fun. 

I've seen it fail quite often for me while running it in isolation:

```
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :only puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service-test/multiple-watch-dirs-test
(WARNING: user-level profile defined in project files.)

Testing puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service-test
timed-out waiting for events

FAIL in (multiple-watch-dirs-test) (filesystem_watch_service_test.clj:301)
Watching of multiple directories using a single watcher Nested watch directories

expected: #{{:path #<java.io.File@57f60b5e /tmp/dir-11469829737696-781167481/nested-dir>,
             :type :create}
            {:path #<java.io.File@21e860cd /tmp/dir-11469829737696-781167481/nested-dir/nested-test-file>,
             :type :create}}
  actual: ({:path #<java.io.File@a739cd9 /tmp/dir-11469829737696-781167481/nested-dir>,
            :type :create}
           {:path #<java.io.File@610f3f1b /tmp/dir-11469829737696-781167481/nested-dir/nested-test-file>,
            :type :modify}
           {:path #<java.io.File@7a7ae0e1 /tmp/dir-11469829737696-781167481/nested-test-file>,
            :type :create})
```

FWIW, when running these tests on pooler VMs I've also seen failure rates, and which tests fail, vary by time of day (assuming they're varying by load on the backend infrastructure).

justinstoller:
The JDK bug seems to be a huge culprit, minus some general raciness of tests that come out when we actually start operating asynchronously. If I'm understanding this correctly the bug occurs when we are registering watchables while also reading watch events. It might be helpful to set some kind of semaphore via atom so that we don't call `.pollEvents` while we are in the middle of a `registerRecursive` call. I need to look at the source code some more to figure out if that's correct but I think that's the route I'm going to investigate next.

justinstoller:
7c902e3 failed on one cell with the JDK issue noted above.

haus:
@justinstoller do you mind adding something to the readme noting the issue and linking to the bug?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,80 +86,32 @@
...
+      (while (not @stopped?)
+        (try
+          (let [watch-key (.take (:watch-service watcher))
+                orig-events (.pollEvents watch-key)
+                events (map #(clojurize % (.watchable watch-key)) orig-events)
+                callbacks @(:callbacks watcher)]
+            (when-not (empty? events)
+              (log/info (trs "Got {0} event(s) for watched-path {1}"
+                             (count orig-events) (.watchable watch-key)))
+              (log/debug (trs "Events:\n{0}"
```
haus: I thought we weren't supposed to put whitespace in a trs call? /cc @janelu2

justinstoller: Yeah, I think you're right I'll go ahead and add a fix for that in this PR.

haus:
Aside from the `trs` comment and adding a link to the README, I think this looks pretty good. I'd like to get @cprice404 's opinion on the sleep in the test, but am otherwise :+1:.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
janelu2: :+1:

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -82,80 +86,32 @@
...
+                              (pprint-events events)))
+              (log/trace (trs "orig-events:\n{0}"
```
haus: (same as above)

cprice404:
I don't have any huge problem with the sleep for now, since a. it's pretty short, and b. these tests are isolated to their own repo so won't cause other projects' suites to run longer, and c. the underlying issue is annoying and stupid and largely out of our hands.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -102,11 +102,13 @@
             (when-not (empty? events)
               (log/info (trs "Got {0} event(s) for watched-path {1}"
                              (count orig-events) (.watchable watch-key)))
-              (log/debug (trs "Events:\n{0}"
-                              (pprint-events events)))
-              (log/trace (trs "orig-events:\n{0}"
-                              (ks/pprint-to-string
-                                (map clojurize-for-logging orig-events))))
+              (log/debug (format "%s\n%s"
```
haus: any reason not to just use `log/debugf` here and do away with the format call? similarly with trace vs tracef below.

justinstoller: Because... I didn't know such a thing existed? I'll use that (;

haus:
will merge once travis says go.

justinstoller:
Three runs in a row that are green!!!!!!

PR #5, opened 2016-07-29
(TK-391) Handle OVERFLOW events

justinstoller:
Previously we made no distinction for incoming overflow events.

OVERFLOW events do not have a specific context associated with their
change - only the watchable that the WatchKey was registered with.
And, an event that lacked a context would cause an NPE when
processing the event.

This patch checks to see if the event in question is an OVERFLOW event
and special cases its reporting to return as much information as
possible, without causing an NPE.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
@@ -6,7 +6,7 @@
   "Schema for an event on a file watched by this service."
   (schema/if #(= (:type %) :unknown)
     {:type (schema/eq :unknown)
-     :path (schema/eq nil)}
+     :path (schema/pred nil?)}
     {:type (schema/enum :create :modify :delete)
      :path File}))
 
--- a/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
@@ -4,8 +4,11 @@
 
 (def Event
   "Schema for an event on a file watched by this service."
-  {:type (schema/enum :create :modify :delete :unknown)
-   :path File})
+  (schema/if #(= (:type %) :unknown)
+    {:type (schema/eq :unknown)
+     :path (schema/eq nil)}
+    {:type (schema/enum :create :modify :delete)
+     :path File}))
 
 (defprotocol Watcher
   (add-watch-dir! [this dir options]
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -31,7 +31,7 @@
    watch-path :- Path]
   {:type (get event-type-mappings (.kind event))
    :path (if (= StandardWatchEventKinds/OVERFLOW (.kind event))
-           (fs/file (.resolve watch-path "UNKNOWN"))
+           nil
            (-> watch-path
                (.resolve (.context event))
                fs/file))})
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -30,8 +30,7 @@
   [event :- WatchEvent
    watch-path :- Path]
   {:type (get event-type-mappings (.kind event))
-   :path (if (= StandardWatchEventKinds/OVERFLOW (.kind event))
-           nil
+   :path (when-not (= StandardWatchEventKinds/OVERFLOW (.kind event))
            (-> watch-path
                (.resolve (.context event))
                fs/file))})
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -30,9 +30,11 @@
   [event :- WatchEvent
    watch-path :- Path]
   {:type (get event-type-mappings (.kind event))
-   :path (-> watch-path
-             (.resolve (.context event))
-             fs/file)})
+   :path (if (= StandardWatchEventKinds/OVERFLOW (.kind event))
+           (fs/file (.resolve watch-path "UNKNOWN"))
+           (-> watch-path
+               (.resolve (.context event))
+               fs/file))})
 
 ;; This is quite similar to the function above but it is more a direct
 ;; conversion of the exact data available on a specific WatchEvent instance,
--- a/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
@@ -387,20 +387,19 @@
             :error))))))
 
 (deftest ^:integration overflow-test
-  (let  [watch-dir  (fs/temp-dir "overflow")
-        results  (atom  [])
-        callback  (fn  [events]  (swap! results concat events))]
+  (let [watch-dir (fs/temp-dir "overflow")
+        results (atom [])
+        callback (make-callback results)]
     (with-app-with-config
-     app watch-service-and-deps  {}
-     (let  [service  (tk-app/get-service app :FilesystemWatchService)]
+     app watch-service-and-deps {}
+     (let [service (tk-app/get-service app :FilesystemWatchService)]
        (watch! service watch-dir callback))
      ;; From experimenting the buffer seems to fill at 512 events.
      ;; However the number of events varies by operating system.
-     (let  [files  (for  [n  (range 1000)]
+     (let [files (for [n (range 1000)]
                    (fs/file watch-dir  (str n ".txt")))]
-       (doseq  [f files]
+       (doseq [f files]
          (spit f "file content"))
       (let [events #{{:type :unknown
-                      :path (fs/file watch-dir "UNKNOWN")}}]
-        (spit (fs/file watch-dir "observed.txt") "file content")
+                      :path nil}}]
         (is (= events (wait-for-events results events))))))))
--- a/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
@@ -385,3 +385,30 @@
        (is (logged?
             #"shutdown-on-error triggered because of exception"
             :error))))))
+
+(deftest ^:integration overflow-test
+  (let  [watch-dir  (fs/temp-dir "overflow")
+        results  (atom  [])
+        callback  (fn  [events]  (swap! results concat events))]
+    (with-app-with-config
+     app watch-service-and-deps  {}
+     (let  [service  (tk-app/get-service app :FilesystemWatchService)]
+       (watch! service watch-dir callback))
+     ;; From experimenting the buffer seems to fill at 512 events.
+     ;; However the number of events varies by operating system.
+     (let  [files  (for  [n  (range 1000)]
+                   (fs/file watch-dir  (str n ".txt")))]
+       (doseq  [f files]
+         (spit f "file content"))
+       (let  [events  (set  (map  (fn  [f]
+                                    {:type :create
+                                     :path f})
+                                  files))]
+          ;; Block for the standard wait period which should be sufficient that
+          ;; any polling interval has been reset and we won't add to an overflow
+          ;; event with our next file creation.
+          (java.lang.Thread/sleep (long wait-time)))
+      (let [events #{{:type :create
+                      :path (fs/file watch-dir "observed.txt")}}]
+        (spit (fs/file watch-dir "observed.txt") "file content")
+        (is (= events (wait-for-events results events))))))))
--- a/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
@@ -400,15 +400,7 @@
                    (fs/file watch-dir  (str n ".txt")))]
        (doseq  [f files]
          (spit f "file content"))
-       (let  [events  (set  (map  (fn  [f]
-                                    {:type :create
-                                     :path f})
-                                  files))]
-          ;; Block for the standard wait period which should be sufficient that
-          ;; any polling interval has been reset and we won't add to an overflow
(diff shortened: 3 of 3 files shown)


justinstoller:
On a linux box this passes though it seems like the tests are generally flaky:

```
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :all

2016-07-28 22:57:40,052 ERROR [p.t.s.s.scheduler-core] scheduled job threw error
java.nio.file.NoSuchFileException: /tmp/single-path-test1469771854015-3706661469/sub-dir
        at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86) ~[na:1.8.0_101]
        at sun.nio.fs.UnixException.asIOException(UnixException.java:111) ~[na:1.8.0_101]
        at sun.nio.fs.LinuxWatchService$Poller.implRegister(LinuxWatchService.java:246) ~[na:1.8.0_101]
        at sun.nio.fs.AbstractPoller.processRequests(AbstractPoller.java:260) ~[na:1.8.0_101]
        at sun.nio.fs.LinuxWatchService$Poller.run(LinuxWatchService.java:329) ~[na:1.8.0_101]
        at java.lang.Thread.run(Thread.java:745) [na:1.8.0_101]
timed-out waiting for events

lein test :only puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service-test/single-path-test

FAIL in (single-path-test) (filesystem_watch_service_test.clj:121)
re-creation of a deleted directory Initial directory creation and deletion
expected: (= events (wait-for-events results events))
  actual: (not (= #{{:path #object[java.io.File 0x1c277413 "/tmp/single-path-test1469771854015-3706661469/sub-dir"], :type :delete}} ({:type :create, :path #object[java.io.File 0x1e9b1d9f "/tmp/single-pat
h-test1469771854015-3706661469/sub-dir"]})))

Ran 7 tests containing 53 assertions.
1 failures, 0 errors.
Tests failed.
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :all

Ran 7 tests containing 53 assertions.
0 failures, 0 errors.
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :all

Ran 7 tests containing 53 assertions.
0 failures, 0 errors.
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :all

timed-out waiting for events

lein test :only puppetlabs.trapperkeeper.services.watcher.filesystem-watch-service-test/nested-files-test

FAIL in (nested-files-test) (filesystem_watch_service_test.clj:247)
new nested directory creation of a file within
expected: (= events (wait-for-events results events))
  actual: (not (= #{{:path #object[java.io.File 0x107302a5 "/tmp/root-dir1469772087919-805879646/another-nested-dir/new-nested-file"], :type :create}} []))

Ran 7 tests containing 53 assertions.
1 failures, 0 errors.
Tests failed.
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :all

Ran 7 tests containing 53 assertions.
0 failures, 0 errors.
[root@d3ql7wtvhgkzwk0 trapperkeeper-filesystem-watcher]# lein test :all

Ran 7 tests containing 53 assertions.
0 failures, 0 errors.
```

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -30,9 +30,11 @@
   [event :- WatchEvent
    watch-path :- Path]
   {:type (get event-type-mappings (.kind event))
-   :path (-> watch-path
-             (.resolve (.context event))
-             fs/file)})
+   :path (if (= StandardWatchEventKinds/OVERFLOW (.kind event))
+           (fs/file (.resolve watch-path "UNKNOWN"))
```
KevinCorcoran: I think this is wrong because it is trying to resolve a file named `UNKNOWN`.  Or, I am not understanding something?

I was thinking we'd need to deal with this by making `:path` either a `schema/optional-key` or `(schema/maybe File)` in the `Event` schema.  Or, something more drastic:

``` clj
(def Event
  "Schema for an event on a file watched by this service."
  (schema/if #(= (:type %) :unknown)
    {:type (schema/eq :unknown)
     :path (schema/maybe File)}
    {:type (schema/enum :create :modify :delete)
     :path File}))
```

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -385,3 +385,22 @@
...
+  (let  [watch-dir  (fs/temp-dir "overflow")
+        results  (atom  [])
+        callback  (fn  [events]  (swap! results concat events))]
+    (with-app-with-config
+     app watch-service-and-deps  {}
+     (let  [service  (tk-app/get-service app :FilesystemWatchService)]
+       (watch! service watch-dir callback))
+     ;; From experimenting the buffer seems to fill at 512 events.
+     ;; However the number of events varies by operating system.
+     (let  [files  (for  [n  (range 1000)]
```
KevinCorcoran: There appear to be some extra spaces on this line (or indentation is just off on the line below, one of the two).

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -385,3 +385,22 @@
...
+(deftest ^:integration overflow-test
+  (let  [watch-dir  (fs/temp-dir "overflow")
+        results  (atom  [])
+        callback  (fn  [events]  (swap! results concat events))]
+    (with-app-with-config
+     app watch-service-and-deps  {}
+     (let  [service  (tk-app/get-service app :FilesystemWatchService)]
+       (watch! service watch-dir callback))
+     ;; From experimenting the buffer seems to fill at 512 events.
+     ;; However the number of events varies by operating system.
```
KevinCorcoran, line 398: If this passes reliably, that's cool.  If it starts to fail intermittently, or we wanted to make it more robust, we could change to create files indefinitely until an overflow event occurs.  But, I'm good with this approach for now.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -385,3 +385,22 @@
        (is (logged?
             #"shutdown-on-error triggered because of exception"
             :error))))))
+
+(deftest ^:integration overflow-test
+  (let  [watch-dir  (fs/temp-dir "overflow")
+        results  (atom  [])
+        callback  (fn  [events]  (swap! results concat events))]
```
KevinCorcoran: This call to `swap!` can be replaced with `(make-callback results)` which is consistent with the other tests.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -385,3 +385,22 @@
...
+       (watch! service watch-dir callback))
+     ;; From experimenting the buffer seems to fill at 512 events.
+     ;; However the number of events varies by operating system.
+     (let  [files  (for  [n  (range 1000)]
+                   (fs/file watch-dir  (str n ".txt")))]
+       (doseq  [f files]
+         (spit f "file content"))
+      (let [events #{{:type :unknown
+                      :path (fs/file watch-dir "UNKNOWN")}}]
+        (spit (fs/file watch-dir "observed.txt") "file content")
```
KevinCorcoran: This looks vestigial, should it be deleted?

KevinCorcoran:
🙈  Only feedback of significant importance was the line where `.resolve` is called with `"UNKNOWN"`.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
justinstoller: That's so weird. I copy pastad your [reproducer](https://github.com/KevinCorcoran/trapperkeeper-filesystem-watcher/commit/53f66ce4df19a47c1c51da30d31899cdf4c934e0) to start this ticket but when I pasted my editor inserted extra spaces everywhere in the test.... I've addressed this in an incoming commit. Apologies.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: We talked about this in standup. This request makes good sense. I've adopted it nearly wholesale.

The only change I made is that now `:path` for an `:unknown` event is `nil` instead of `maybe` `File`.

In the current implementation we know that `:unknown` will always have a `nil` `:path`. Alternatively, if we did something more like:

``` clj
  :path (if-let [context (.context event)]
          (-> watch-path
            (.resolve context)
            fs/file)
          nil)
```

Then we wouldn't necessarily know when `:path` was `nil` and we could so something like this:

``` clj
(def Event
  "Schema for an event on a file watched by this service."
  {:type (schema/enum :create :modify :delete :unknown)
   :path (schema/maybe File)})
```

The current impl in this PR is more brittle in the case that a normal event could return a null context, but I think the current impl more explicitly describes our (or at least my) current understanding of how the world works.

justinstoller:
Addressed PR feedback, love to have @haus take a look at it!

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -30,9 +30,11 @@
   [event :- WatchEvent
    watch-path :- Path]
   {:type (get event-type-mappings (.kind event))
-   :path (-> watch-path
-             (.resolve (.context event))
-             fs/file)})
+   :path (if (= StandardWatchEventKinds/OVERFLOW (.kind event))
```
haus: if you make this a `when-not` instead of an if, you can remove the nil branch as it will eval to nil in the case of an overflow.

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
```diff
@@ -4,8 +4,11 @@
 
 (def Event
   "Schema for an event on a file watched by this service."
-  {:type (schema/enum :create :modify :delete :unknown)
-   :path File})
+  (schema/if #(= (:type %) :unknown)
+    {:type (schema/eq :unknown)
+     :path (schema/eq nil)}
```
haus: I don't think I've seen this usage before. I think it's more common to do `(schema/pred nil?)`, but this is equivalent in function, so I don't think it matters.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -385,3 +385,21 @@
...
+  (let [watch-dir (fs/temp-dir "overflow")
+        results (atom [])
+        callback (make-callback results)]
+    (with-app-with-config
+     app watch-service-and-deps {}
+     (let [service (tk-app/get-service app :FilesystemWatchService)]
+       (watch! service watch-dir callback))
+     ;; From experimenting the buffer seems to fill at 512 events.
+     ;; However the number of events varies by operating system.
+     (let [files (for [n (range 1000)]
```
haus: I think this can be more concisely written with `dotimes`.

``` clojure
(dotimes [n 1000]
  (spit (format "%s/%s.txt" watch-dir n) "file content"))
```

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -385,3 +385,21 @@
...
+                   (fs/file watch-dir  (str n ".txt")))]
+       (doseq [f files]
+         (spit f "file content"))
+      (let [events #{{:type :unknown
+                      :path nil}}]
+        (is (= events (wait-for-events results events))))))))
```
haus: It seems possible we will get a few real events as well as an overflow here?

Is equality or containment the assertion we care about? Could the `contains-events?` function be used here instead of strict equality?

Edit: Ooops. And I misread the `contains-events?` function. Never mind me. I'm still interested in the first question though.

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
justinstoller: Cool, I didn't find a good example of ensuring `nil` (lots of `schema/maybe`) with a quick skim. Happy to change that. Could you link me to an example?

haus: Yea, I can't find any examples, so maybe I made it up. `schema/pred` is more common than `schema/eq` in general, but `(= nil thing)` and `(nil? thing)` should be equivalent, so :shrug:.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
justinstoller: I thought that as well, but from my experiments on Linux it appears that we do not.

We will get normal events until we overflow the buffer at which point we get _an_ overflow event that begins incrementing a counter of how many _additional_ events occurred until we reset the key.

Here is the log output from running this test at trace:

```
2016-08-01 11:28:56,303 INFO  [p.t.s.w.filesystem-watch-core] Got 1 event(s) for watched-path /tmp/overflow1470076135277-2767012267
2016-08-01 11:28:56,304 DEBUG [p.t.s.w.filesystem-watch-core] Events:
({:type :unknown, :path ""})

2016-08-01 11:28:56,306 TRACE [p.t.s.w.filesystem-watch-core] orig-events:
({:context nil,
  :count 1488,
  :kind
  #object[java.nio.file.StandardWatchEventKinds$StdWatchEventKind 0x49c745cd "OVERFLOW"]})
```

justinstoller:
I think this is ready to go @haus

haus:
Okay, I'm good here. :+1:

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
```diff
@@ -4,8 +4,11 @@
 
 (def Event
   "Schema for an event on a file watched by this service."
-  {:type (schema/enum :create :modify :delete :unknown)
-   :path File})
+  (schema/if #(= (:type %) :unknown)
+    {:type (schema/eq :unknown)
+     :path (schema/pred nil?)}
```
KevinCorcoran, line 9: `(schema/pred nil?)` seems very strange to me, I guess that matches any value or `nil`, huh?  I've used `Object` for that in the past but I'm not sure if either one is better.  The best solution seems like it would probably be to just omit the `:path` key for `:unknown` events.

PR #6, opened 2016-08-02
(TK-391) Add unit test for overflow event handling

justinstoller:
Previously we tested overflow handling at the :integration level. This
has the unfortunate quality of requiring us to cause an error state
within Java/inotify. Up until this point we have been able to do this
easily without major repercussions on the systems that run the test.

That is not true in CI however, where we have a version of the JDK that
will not easily throw overflow events. In fact going to the lengths
required to cause an overflow event on these hosts is just as likely to
crash the JVM or fill up the disk.

This patch adds a unit test to prove our handling of overflow events is
satisfactory, and refactors the production code to make it unit
testable.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -6,7 +6,7 @@
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.trapperkeeper.services.protocols.filesystem-watch-service :refer [Event Watcher]])
   (:import (clojure.lang Atom IFn)
-           (java.nio.file StandardWatchEventKinds Path WatchEvent FileSystems ClosedWatchServiceException)
+           (java.nio.file StandardWatchEventKinds Path WatchEvent WatchKey FileSystems ClosedWatchServiceException)
            (com.puppetlabs DirWatchUtils)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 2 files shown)


justinstoller:
This should fix our current issues in CI

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -395,20 +396,25 @@
...
+                    watch-path
+                    (:watch-service watcher)
+                    (into-array [java.nio.file.StandardWatchEventKinds/ENTRY_CREATE]))
+        events [overflow-event]
+        actual (atom [])
+        expected #{{:type :unknown :path nil}}
+        callback (make-callback actual)]
+    (with-test-logging
+      (add-callback! watcher callback)
+      (watch-core/process-events watcher watch-key events #(%))
```
haus: I think making this `(fn [func] (func))` might make it a little more readable.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -395,20 +396,25 @@
...
-     ;; However the number of events varies by operating system.
-     ;; Once the buffer overflows we cannot reliably receive any other
-     ;; kind of event.
-     (dotimes [n 1000]
-       (spit (format "%s/%s.txt" watch-dir n) "file content"))
-     (let [events #{{:type :unknown
-                     :path nil}}]
-       (is (= events (wait-for-events results events)))))))
+(def overflow-event
+  (reify
```
haus, line 403: Would be good to document that this is implementing an interface that is otherwise private to us (as Event is a private static class inside of an abstract class) https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/sun/nio/fs/AbstractWatchKey.java#L190-L222

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,33 +87,42 @@
                                           (filter dir-create?)
                                           (map #(.toPath (:path %)))))))
 
+(defn retrieve-events
```
haus: Having a schema here would be cool.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,33 +87,42 @@
...
+  [watcher]
+  (let [watch-key (.take (:watch-service watcher))
+        events (.pollEvents watch-key)]
+    [watch-key events]))
+
+(defn process-events
```
haus: Same as above, using a schema would make it clear what is going on.

haus: Also, this might be better called `process-events!` as it is all about the side effects.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,33 +87,42 @@
...
+  [watcher watch-key java-events shutdown-fn]
+  (let [clojure-events (map #(clojurize % (.watchable watch-key)) java-events)
+        callbacks @(:callbacks watcher)]
+    (log/info (trs "Got {0} event(s) for watched-path {1}"
+                   (count java-events) (.watchable watch-key)))
+    (log/debugf "%s\n%s"
+                (trs "Events:")
+                (pprint-events clojure-events))
+    (log/tracef "%s\n%s"
+                (trs "orig-events:")
```
haus, line 113: Would be good to settle on orig-events or java-events and just have one. Having both means trying to figure out which clojure var the "orig-events" string is referring to. I'd vote to stick with orig-events, as users seeing logs probably aren't going to care about the java part of the event as much as that it is a preprocessed event.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -395,20 +396,25 @@
...
+    java.nio.file.WatchEvent
+    (kind [this] java.nio.file.StandardWatchEventKinds/OVERFLOW)
+    (count [this] 1)
+    (context [this] nil)))
+
+(deftest process-events-handles-overflows-correctly
```
haus: Having at least one `testing` call here to describe what is going on would be good.

haus:
This looks pretty good. I left some schema/naming/documentation comments, but once those are resolved I'm good merging this.

justinstoller:
D'oh! I was so excited at getting the damned thing to work I opened the PR without going back and updating the fns with schemas and docs. I believe I've addressed all your feedback, @haus

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,33 +87,48 @@
                                           (filter dir-create?)
                                           (map #(.toPath (:path %)))))))
 
+(schema/defn retrieve-events
```
haus, line 90: Would it make sense to add a schema here for return type?

haus:
Alright, I'm in. :+1:

PR #8, opened 2016-08-05
(TK-389) Debounce Events

justinstoller:
This patch set follows `Implementation Option #2` in the corresponding ticket.

After an event occurs it begins a polling loop to see if there are queued events that occurred concurrently to the one being processed. Once the queue has been drained it will wait until `window-min` (currently set to 100 milliseconds) to see if any additional events occur. It will break and report the events that it has accumulated thus far if `window-max` (currently set to 2 seconds) is exceeded during its polling loop.

This PR contains two additional refactors. It consolidates our internal representation of Events so that we have one collection to represent them that all internal functions can use (and has appropriate helpers to translate them so that consumers do not need to change). It also moves the TK shutdown service function to wrap the entire event processing rather than just the consumer specified callback. This allows any unforeseen issues (eg schema errors) to be logged and raised when they would have normally failed silently in the processing future.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
@@ -6,9 +6,15 @@
   "Schema for an event on a file watched by this service."
   (schema/if #(= (:type %) :unknown)
     {:type (schema/eq :unknown)
-     :path (schema/pred nil?)}
+     :count schema/Int
+     :watched-path File
+     :changed-path (schema/pred nil?)
+     :full-path (schema/pred nil?)}
     {:type (schema/enum :create :modify :delete)
-     :path File}))
+     :count schema/Int
+     :watched-path File
+     :changed-path File
+     :full-path File}))
 
 (defprotocol Watcher
   (add-watch-dir! [this dir options]
--- a/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
@@ -7,14 +7,11 @@
   (schema/if #(= (:type %) :unknown)
     {:type (schema/eq :unknown)
      :count schema/Int
-     :watched-path File
-     :changed-path (schema/pred nil?)
-     :full-path (schema/pred nil?)}
+     :watched-path File}
     {:type (schema/enum :create :modify :delete)
      :count schema/Int
      :watched-path File
-     :changed-path File
-     :full-path File}))
+     :changed-path File}))
 
 (defprotocol Watcher
   (add-watch-dir! [this dir options]
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -34,33 +34,23 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (schema/defn clojurize :- Event
+  "Takes the Java WatchEvent and the watchable Java Path that the event
+  occurred within and creates a Clojure map to represent the Event
+  throughout the system. The watched-path key is the registered watchable,
+  the changed-path the relative path to what changed, and the full path is
+  the absolute path to the changed event (watched-path + changed-path)."
   [event :- WatchEvent
    watched-path :- Path]
   (let [kind (get event-type-mappings (.kind event))
-        changed-path (when-not (= :unknown kind)
-                       (.context event))
-        full-path (when-not (= :unknown kind)
-                    (-> watched-path (.resolve changed-path) (.toFile)))]
-    {:type kind
-     :count (.count event)
-     :watched-path (.toFile watched-path)
-     :changed-path (when changed-path (.toFile changed-path))
-     :full-path full-path}))
-
-(schema/defn format-for-debugging
-  [{:keys [changed-path count type]} :- Event]
-  {:context changed-path
-   :count count
-   :kind type})
-
-(schema/defn format-for-consumers
-  [{:keys [full-path type]} :- Event]
-  {:path full-path
-   :type type})
-
-(schema/defn format-for-users
-  [event :- Event]
-  (update (format-for-consumers event) :path str))
+        count (.count event)]
+    (if (= :unknown kind)
+      {:type kind
+       :count count
+       :watched-path (.toFile watched-path)}
+      {:type kind
+       :count count
+       :watched-path (.toFile watched-path)
+       :changed-path (.. watched-path (resolve (.context event)) (toFile))})))
 
 (defn validate-watch-options!
   [options]
@@ -92,56 +82,55 @@
    watcher :- (schema/protocol Watcher)]
   (let [dir-create? (fn [event]
                       (and (= :create (:type event))
-                           (fs/directory? (:full-path event))))]
+                           (fs/directory? (:changed-path event))))]
     (DirWatchUtils/registerRecursive (:watch-service watcher)
                                      (->> events
                                           (filter dir-create?)
-                                          (map #(.toPath (:full-path %)))))))
+                                          (map #(.toPath (:changed-path %)))))))
 
-(defn get-event-maps-from-key
-  [watch-key]
+(schema/defn watch-key->events :- [Event]
+  [watch-key :- WatchKey]
   (let [events (.pollEvents watch-key)]
     (map #(clojurize % (.watchable watch-key)) events)))
 
 (schema/defn retrieve-events :- [Event]
   "Blocks until an event the watcher is concerned with has occured. Will then
-  poll for a new event, watiting at least `window-min` for a new event to
+  poll for a new event, waiting at least `window-min` for a new event to
   occur. Will continue polling for as long as there are new events that occur
   within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (get-event-maps-from-key watch-key)
+        initial-events (watch-key->events watch-key)
         time-limit (+ (System/currentTimeMillis) window-max)]
+    (watch-new-directories! initial-events watcher)
     (.reset watch-key)
-    (watch-new-directories! events watcher)
-      (loop [events' events]
+    (if-not (empty? initial-events)
+      (loop [events initial-events]
         (if-let [waiting-key (.poll (:watch-service watcher) window-min window-units)]
-          (let [waiting-events (get-event-maps-from-key waiting-key)]
+          (let [waiting-events (watch-key->events waiting-key)]
             (watch-new-directories! waiting-events watcher)
             (.reset waiting-key)
             (if (< (System/currentTimeMillis) time-limit)
-              (recur (into events' waiting-events))
-              (into events' waiting-events)))
-          events'))))
+              (recur (concat events waiting-events))
+              (concat events waiting-events)))
+          events))
+      initial-events)))
 
 
 (schema/defn process-events!
   "Process for side-effects any events that occured for watcher's watch-key"
   [watcher :- (schema/protocol Watcher)
    events :- [Event]]
-  (let [callbacks @(:callbacks watcher)]
-    (log/info (trs "Got {0} event(s) on path(s) {1}"
-                   (count events) (distinct (map #(:full-path %) events))))
-    (log/debugf "%s\n%s"
-                (trs "Events:")
-                (ks/pprint-to-string
-                  (map format-for-users events)))
+  (let [callbacks @(:callbacks watcher)
+        events-by-dir (group-by :watched-path events)]
+    (doseq [[dir events'] events-by-dir]
+      (log/info (trs "Got {0} event(s) in directory {1}"
+                   (count events') dir)))
     (log/tracef "%s\n%s"
-                (trs "orig-events:")
-                (ks/pprint-to-string
-                  (map format-for-debugging events)))
+                (trs "Events:")
+                (ks/pprint-to-string events))
     (doseq [callback callbacks]
-      (callback (map format-for-consumers events)))))
+      (callback events))))
 
 (schema/defn watch!
   "Creates a future and processes events for the passed in watcher.
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -101,8 +101,7 @@
   "Process for side-effects any events that occured for watcher's watch-key"
   [watcher :- (schema/protocol Watcher)
    watch-key :- WatchKey
-   orig-events :- [WatchEvent]
-   shutdown-fn :- IFn]
+   orig-events :- [WatchEvent]]
   (let [clojure-events (map #(clojurize % (.watchable watch-key)) orig-events)
         callbacks @(:callbacks watcher)]
     (log/info (trs "Got {0} event(s) for watched-path {1}"
@@ -114,8 +113,8 @@
                 (trs "orig-events:")
                 (ks/pprint-to-string
                   (map clojurize-for-logging orig-events)))
-    (shutdown-fn #(doseq [callback callbacks]
-                   (callback clojure-events)))
+    (doseq [callback callbacks]
+      (callback clojure-events))
     (watch-new-directories! clojure-events watcher)))
 
 (schema/defn watch!
@@ -125,12 +124,12 @@
    shutdown-fn :- IFn]
   (future
     (let [stopped? (atom false)]
-      (while (not @stopped?)
-        (try
-          (let [[watch-key events] (retrieve-events watcher)]
-            (when-not (empty? events)
-              (process-events! watcher watch-key events shutdown-fn)))
-         (catch ClosedWatchServiceException e
-           (reset! stopped? true)
-           (log/info (trs "Closing watcher {0}" watcher))))))))
+      (shutdown-fn #(while (not @stopped?)
+                     (try
+                       (let [[watch-key events] (retrieve-events watcher)]
+                         (when-not (empty? events)
+                           (process-events! watcher watch-key events)))
+                      (catch ClosedWatchServiceException e
+                        (reset! stopped? true)
+                        (log/info (trs "Closing watcher {0}" watcher)))))))))
 
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -40,11 +40,11 @@
         changed-path (when-not (= :unknown kind)
                        (.context event))
         full-path (when-not (= :unknown kind)
-                    (-> watched-path (.resolve changed-path) fs/file))]
+                    (-> watched-path (.resolve changed-path) (.toFile)))]
     {:type kind
      :count (.count event)
-     :watched-path (fs/file watched-path)
-     :changed-path (fs/file changed-path)
+     :watched-path (.toFile watched-path)
+     :changed-path (when changed-path (.toFile changed-path))
      :full-path full-path}))
 
 (schema/defn format-for-debugging
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -95,6 +95,7 @@
   (let [watch-key (.take (:watch-service watcher))
         events (.pollEvents watch-key)]
     (.reset watch-key)
+    (watch-new-directories! (map #(clojurize % (.watchable watch-key)) events) watcher)
     [watch-key events]))
 
 (schema/defn process-events!
@@ -114,8 +115,7 @@
                 (ks/pprint-to-string
                   (map clojurize-for-logging orig-events)))
     (doseq [callback callbacks]
-      (callback clojure-events))
-    (watch-new-directories! clojure-events watcher)))
+      (callback clojure-events))))
 
 (schema/defn watch!
   "Creates a future and processes events for the passed in watcher.
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -35,27 +35,32 @@
 
 (schema/defn clojurize :- Event
   [event :- WatchEvent
-   watch-path :- Path]
-  {:type (get event-type-mappings (.kind event))
-   :path (when-not (= StandardWatchEventKinds/OVERFLOW (.kind event))
-           (-> watch-path
-               (.resolve (.context event))
-               fs/file))})
+   watched-path :- Path]
+  (let [kind (get event-type-mappings (.kind event))
+        changed-path (when-not (= :unknown kind)
+                       (.context event))
+        full-path (when-not (= :unknown kind)
+                    (-> watched-path (.resolve changed-path) fs/file))]
+    {:type kind
+     :count (.count event)
+     :watched-path (fs/file watched-path)
+     :changed-path (fs/file changed-path)
+     :full-path full-path}))
 
-;; This is quite similar to the function above but it is more a direct
-;; conversion of the exact data available on a specific WatchEvent instance,
-;; only used for debugging.
-(defn clojurize-for-logging
-  [e]
-  {:context (.context e)
-   :count (.count e)
-   :kind (.kind e)})
+(schema/defn format-for-debugging
+  [{:keys [changed-path count type]} :- Event]
+  {:context changed-path
+   :count count
+   :kind type})
 
-(defn pprint-events
-  [events]
-  (->> events
-       (map #(update % :path str))
-       ks/pprint-to-string))
+(schema/defn format-for-consumers
+  [{:keys [full-path type]} :- Event]
+  {:path full-path
+   :type type})
+
+(schema/defn format-for-users
+  [event :- Event]
+  (update (format-for-consumers event) :path str))
 
 (defn validate-watch-options!
   [options]
@@ -87,55 +92,56 @@
    watcher :- (schema/protocol Watcher)]
   (let [dir-create? (fn [event]
                       (and (= :create (:type event))
-                           (fs/directory? (:path event))))]
+                           (fs/directory? (:full-path event))))]
     (DirWatchUtils/registerRecursive (:watch-service watcher)
                                      (->> events
                                           (filter dir-create?)
-                                          (map #(.toPath (:path %)))))))
+                                          (map #(.toPath (:full-path %)))))))
 
-(schema/defn retrieve-events :- {WatchKey [WatchEvent]}
+(defn get-event-maps-from-key
+  [watch-key]
+  (let [events (.pollEvents watch-key)]
+    (map #(clojurize % (.watchable watch-key)) events)))
+
+(schema/defn retrieve-events :- [Event]
   "Blocks until an event the watcher is concerned with has occured. Will then
   poll for a new event, watiting at least `window-min` for a new event to
   occur. Will continue polling for as long as there are new events that occur
   within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (vec (.pollEvents watch-key))
-        time-limit (+ (System/currentTimeMillis) window-max)
-        events-by-key {watch-key events}]
+        events (get-event-maps-from-key watch-key)
+        time-limit (+ (System/currentTimeMillis) window-max)]
     (.reset watch-key)
-    (watch-new-directories! (map #(clojurize % (.watchable watch-key)) events) watcher)
-      (loop [keys-events events-by-key]
+    (watch-new-directories! events watcher)
+      (loop [events' events]
         (if-let [waiting-key (.poll (:watch-service watcher) window-min window-units)]
-          (let [waiting-events (vec (.pollEvents waiting-key))]
-            (watch-new-directories! (map #(clojurize % (.watchable waiting-key)) waiting-events) watcher)
+          (let [waiting-events (get-event-maps-from-key waiting-key)]
+            (watch-new-directories! waiting-events watcher)
             (.reset waiting-key)
             (if (< (System/currentTimeMillis) time-limit)
-              (recur (update keys-events waiting-key into waiting-events))
-              (update keys-events waiting-key waiting-events)))
-          keys-events))))
+              (recur (into events' waiting-events))
+              (into events' waiting-events)))
+          events'))))
 
 
 (schema/defn process-events!
   "Process for side-effects any events that occured for watcher's watch-key"
   [watcher :- (schema/protocol Watcher)
-   events-by-key :- {WatchKey [WatchEvent]}]
-  (let [orig-events (flatten (vals events-by-key))
-        clojure-events (mapcat
-                         (fn [kv] (map #(clojurize % (.watchable (first kv))) (second kv)))
-                         events-by-key)
-        callbacks @(:callbacks watcher)]
+   events :- [Event]]
+  (let [callbacks @(:callbacks watcher)]
     (log/info (trs "Got {0} event(s) on path(s) {1}"
-                   (count orig-events) (map #(.watchable %) (keys events-by-key))))
+                   (count events) (distinct (map #(:full-path %) events))))
     (log/debugf "%s\n%s"
                 (trs "Events:")
-                (pprint-events clojure-events))
+                (ks/pprint-to-string
+                  (map format-for-users events)))
     (log/tracef "%s\n%s"
                 (trs "orig-events:")
                 (ks/pprint-to-string
-                  (map clojurize-for-logging orig-events)))
+                  (map format-for-debugging events)))
     (doseq [callback callbacks]
-      (callback clojure-events))))
+      (callback (map format-for-consumers events)))))
 
 (schema/defn watch!
   "Creates a future and processes events for the passed in watcher.
@@ -146,9 +152,9 @@
     (let [stopped? (atom false)]
       (shutdown-fn #(while (not @stopped?)
                      (try
-                       (let [events-by-key (retrieve-events watcher)]
-                         (when-not (empty? (vals events-by-key))
-                           (process-events! watcher events-by-key)))
+                       (let [events (retrieve-events watcher)]
+                         (when-not (empty? events)
+                           (process-events! watcher events)))
                       (catch ClosedWatchServiceException e
                         (reset! stopped? true)
                         (log/info (trs "Closing watcher {0}" watcher)))))))))
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -19,6 +19,12 @@
    StandardWatchEventKinds/ENTRY_DELETE :delete
    StandardWatchEventKinds/OVERFLOW :unknown})
 
+(def window-min 100)
+
+(def window-max 2000)
+
+(def window-units java.util.concurrent.TimeUnit/MILLISECONDS)
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Functions
 ;;;
@@ -87,26 +93,40 @@
                                           (filter dir-create?)
                                           (map #(.toPath (:path %)))))))
 
-(schema/defn retrieve-events
-  :- [(schema/one WatchKey "key") (schema/one [WatchEvent] "events")]
-  "Blocks until an event the watcher is concerned with has occured.
-  Returns the native WatchKey and WatchEvents"
+(schema/defn retrieve-events :- {WatchKey [WatchEvent]}
+  "Blocks until an event the watcher is concerned with has occured. Will then
+  poll for a new event, watiting at least `window-min` for a new event to
+  occur. Will continue polling for as long as there are new events that occur
+  within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (.pollEvents watch-key)]
+        events (vec (.pollEvents watch-key))
+        time-limit (+ (System/currentTimeMillis) window-max)
+        events-by-key {watch-key events}]
     (.reset watch-key)
     (watch-new-directories! (map #(clojurize % (.watchable watch-key)) events) watcher)
-    [watch-key events]))
+      (loop [keys-events events-by-key]
+        (if-let [waiting-key (.poll (:watch-service watcher) window-min window-units)]
+          (let [waiting-events (vec (.pollEvents waiting-key))]
+            (watch-new-directories! (map #(clojurize % (.watchable waiting-key)) waiting-events) watcher)
+            (.reset waiting-key)
+            (if (< (System/currentTimeMillis) time-limit)
+              (recur (update keys-events waiting-key into waiting-events))
+              (update keys-events waiting-key waiting-events)))
+          keys-events))))
+
 
 (schema/defn process-events!
   "Process for side-effects any events that occured for watcher's watch-key"
   [watcher :- (schema/protocol Watcher)
-   watch-key :- WatchKey
-   orig-events :- [WatchEvent]]
-  (let [clojure-events (map #(clojurize % (.watchable watch-key)) orig-events)
+   events-by-key :- {WatchKey [WatchEvent]}]
+  (let [orig-events (flatten (vals events-by-key))
+        clojure-events (mapcat
+                         (fn [kv] (map #(clojurize % (.watchable (first kv))) (second kv)))
+                         events-by-key)
         callbacks @(:callbacks watcher)]
-    (log/info (trs "Got {0} event(s) for watched-path {1}"
-                   (count orig-events) (.watchable watch-key)))
+    (log/info (trs "Got {0} event(s) on path(s) {1}"
+                   (count orig-events) (map #(.watchable %) (keys events-by-key))))
     (log/debugf "%s\n%s"
                 (trs "Events:")
                 (pprint-events clojure-events))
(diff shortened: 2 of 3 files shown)


justinstoller:
I don't believe there's a way to test this without creating something race prone. I'm still inclined to do so, but I wanted to get feedback on the implementation so far.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -95,6 +95,7 @@
   (let [watch-key (.take (:watch-service watcher))
         events (.pollEvents watch-key)]
     (.reset watch-key)
+    (watch-new-directories! (map #(clojurize % (.watchable watch-key)) events) watcher)
```
haus: This same call `(map #(clojurize % (.watchable watch-key)) events)` happens elsewhere. Would it be worth breaking into a utility function?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,26 +93,40 @@
                                           (filter dir-create?)
                                           (map #(.toPath (:path %)))))))
 
-(schema/defn retrieve-events
-  :- [(schema/one WatchKey "key") (schema/one [WatchEvent] "events")]
-  "Blocks until an event the watcher is concerned with has occured.
-  Returns the native WatchKey and WatchEvents"
+(schema/defn retrieve-events :- {WatchKey [WatchEvent]}
+  "Blocks until an event the watcher is concerned with has occured. Will then
+  poll for a new event, watiting at least `window-min` for a new event to
```
haus: watiting => waiting

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,26 +93,40 @@
...
+  occur. Will continue polling for as long as there are new events that occur
+  within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (.pollEvents watch-key)]
+        events (vec (.pollEvents watch-key))
```
haus: Why cast the events to a vector?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,26 +93,40 @@
...
+        time-limit (+ (System/currentTimeMillis) window-max)
+        events-by-key {watch-key events}]
     (.reset watch-key)
     (watch-new-directories! (map #(clojurize % (.watchable watch-key)) events) watcher)
-    [watch-key events]))
+      (loop [keys-events events-by-key]
+        (if-let [waiting-key (.poll (:watch-service watcher) window-min window-units)]
+          (let [waiting-events (vec (.pollEvents waiting-key))]
```
haus: why coerce the events into a vector?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,26 +93,40 @@
...
-  (let [clojure-events (map #(clojurize % (.watchable watch-key)) orig-events)
+   events-by-key :- {WatchKey [WatchEvent]}]
+  (let [orig-events (flatten (vals events-by-key))
+        clojure-events (mapcat
+                         (fn [kv] (map #(clojurize % (.watchable (first kv))) (second kv)))
+                         events-by-key)
         callbacks @(:callbacks watcher)]
-    (log/info (trs "Got {0} event(s) for watched-path {1}"
-                   (count orig-events) (.watchable watch-key)))
+    (log/info (trs "Got {0} event(s) on path(s) {1}"
```
haus: This might be more readable if it were broken down per path. I worry it will be really hard to parse with a lot of events happening over many directories.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -126,9 +146,9 @@
     (let [stopped? (atom false)]
       (shutdown-fn #(while (not @stopped?)
                      (try
-                       (let [[watch-key events] (retrieve-events watcher)]
-                         (when-not (empty? events)
-                           (process-events! watcher watch-key events)))
+                       (let [events-by-key (retrieve-events watcher)]
+                         (when-not (empty? (vals events-by-key))
```
haus: What will `events-by-key` be when it has no vals? Could you have a key in events-by-key without any values?

Reading the `retrieve-events` function, I'm not seeing a way for it to return with an empty return, as it blocks on the take until there are events to report.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -35,27 +35,32 @@
...
+                    (-> watched-path (.resolve changed-path) fs/file))]
+    {:type kind
+     :count (.count event)
+     :watched-path (fs/file watched-path)
+     :changed-path (fs/file changed-path)
+     :full-path full-path}))
+
+(schema/defn format-for-debugging
+  [{:keys [changed-path count type]} :- Event]
+  {:context changed-path
```
haus: Does changing the key name help in debugging?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,55 +92,56 @@
...
     (DirWatchUtils/registerRecursive (:watch-service watcher)
                                      (->> events
                                           (filter dir-create?)
-                                          (map #(.toPath (:path %)))))))
+                                          (map #(.toPath (:full-path %)))))))
 
-(schema/defn retrieve-events :- {WatchKey [WatchEvent]}
+(defn get-event-maps-from-key
+  [watch-key]
+  (let [events (.pollEvents watch-key)]
```
haus, line 93: Should this be `if-let` or do we want to always map even if there are no events?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -87,55 +92,56 @@
...
-                           (fs/directory? (:path event))))]
+                           (fs/directory? (:full-path event))))]
     (DirWatchUtils/registerRecursive (:watch-service watcher)
                                      (->> events
                                           (filter dir-create?)
-                                          (map #(.toPath (:path %)))))))
+                                          (map #(.toPath (:full-path %)))))))
 
-(schema/defn retrieve-events :- {WatchKey [WatchEvent]}
+(defn get-event-maps-from-key
```
haus: A schema here would be cool.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -125,12 +150,12 @@
...
-          (let [[watch-key events] (retrieve-events watcher)]
-            (when-not (empty? events)
-              (process-events! watcher watch-key events shutdown-fn)))
-         (catch ClosedWatchServiceException e
-           (reset! stopped? true)
-           (log/info (trs "Closing watcher {0}" watcher))))))))
+      (shutdown-fn #(while (not @stopped?)
+                     (try
+                       (let [events (retrieve-events watcher)]
+                         (when-not (empty? events)
```
haus, line 145: Is there actually a case where `retrieve-events` will return with an empty seq of events?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +92,56 @@
...
+  (let [events (.pollEvents watch-key)]
+    (map #(clojurize % (.watchable watch-key)) events)))
 
-(schema/defn retrieve-events
-  :- [(schema/one WatchKey "key") (schema/one [WatchEvent] "events")]
-  "Blocks until an event the watcher is concerned with has occured.
-  Returns the native WatchKey and WatchEvents"
+(schema/defn retrieve-events :- [Event]
+  "Blocks until an event the watcher is concerned with has occured. Will then
+  poll for a new event, watiting at least `window-min` for a new event to
```
haus: watiting => waiting

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller, line 145: The Oracle docs[1] say that `.pollEvents` could return an empty array. I think if that's the case then `get-events-from-watch-key` will return an empty seq, and if there's no additional events gathered in the loop that empty seq will be returned from `retrieve-events`.

I don't believe that that should happen unless we called `.pollEvents` on a `WatchKey` that hadn't been enqueued in the `WatchService`, but I haven't found docs that say definitively one way or the other. I should caveat that with I haven't looked too hard as that basic flow has existed from the initial implementation of this service.
1. [java.nio.file.WatchKey#pollEvents()](https://docs.oracle.com/javase/8/docs/api/java/nio/file/WatchKey.html#pollEvents--)

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: I found `changed-path` more intuitive than `context` which is what it was prior to this PR.

`context` is the accessor's name on the Java object, this format was originally meant (if I'm reading the tea leaves correctly) to basically be a serialization of the WatchEvent interface for those familiar with it.

I've tried to consolidate the structures we pass around internally while retaining the same output to our consumers.

I changed `context` to `changed-path` because using the closest-names-to-java-members/what-was-there-before would have yielded something like: `watchable`, `context`, and `path`, which I thought were more intuitively referred to as `watched-path`, `changed-path`, and `full-path` (I mainly thought that `path` should be disambiguated).

I'm happy to either refer to them with the former listed keys, or to keep the latter keys and not translate them back. Do you have a preference?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -29,27 +35,32 @@
...
+   watched-path :- Path]
+  (let [kind (get event-type-mappings (.kind event))
+        changed-path (when-not (= :unknown kind)
+                       (.context event))
+        full-path (when-not (= :unknown kind)
+                    (-> watched-path (.resolve changed-path) fs/file))]
+    {:type kind
+     :count (.count event)
+     :watched-path (fs/file watched-path)
+     :changed-path (fs/file changed-path)
```
justinstoller: This is returning the wrong pathed `File`. This is an err on my part. I wasn't aware that [`me.raynes.fs/file`](http://raynes.github.io/fs/me.raynes.fs.html#var-file) always roots relative files in the CWD which is the wrong behavior (we want it to continue to be a relative path).

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
haus: I don't care too much, I mostly was thinking that if the keys were named what we want to start with, you could just call `select-keys` on the elements of the map you care about instead of munging it to a new map.

haus:
Nothing scary here. 🙈 

Would like to get another set of 👀  on it before merging.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller, line 93: My understanding was that `if-let` would still execute if the array was empty. We may get an empty array here, but we should never get, nor return, nil. We could short circuit the mapping if the return value was empty, but I assumed that clojure/jvm would optimize that. I definitely can if you think it's valuable.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: This is still relevant and I went back and forth on it myself. If you think so I'm happy to partition the events by path, `doseq` over the groups and log for each.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
haus, line 93: Oh right, you get an empty array, not a nil. Ignore me.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
haus: Having an example of what this looks like with a few events across directories would be cool to look at and assess.

justinstoller: Here's an example of the current intended output†, broken at my screen width, otherwise it would all be on one line:

```
2016-08-06 09:27:48,665 INFO  [p.t.s.w.filesystem-watch-core] Got 38 event(s) on path(s) ("/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/38.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/39.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/40.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/41.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/42.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/43.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/44.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/45.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/46.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/47.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/48.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/49.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/50.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/51.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/52.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/53.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/54.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/55.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/56.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/57.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/58.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/59.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/60.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/61.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/62.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/63.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/64.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/65.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/66.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/67.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/68.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/69.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/70.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/71.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/72.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/73.txt" "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/74.txt" "/var/folder
s/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470500862356-1130756734/window-max/75.txt")
```

Spending a bit of time playing with the logging, here's an example of just showing events per `watched-path`. I've included some of the debug output below as well:

```
2016-08-06 09:38:31,983 INFO  [p.t.s.w.filesystem-watch-core] Got 38 event(s) in directory /var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470501507670-1186788113/window-max
2016-08-06 09:38:32,087 DEBUG [p.t.s.w.filesystem-watch-core] Events:
({:path
  "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470501507670-1186788113/window-max/0.txt",
  :type :create}
 {:path
  "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470501507670-1186788113/window-max/1.txt",
  :type :create}
 {:path
  "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470501507670-1186788113/window-max/10.txt",
  :type :create}
 {:path
  "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470501507670-1186788113/window-max/11.txt",
  :type :create}
 {:path
  "/var/folders/4l/f8knljcx4pl4sh14txnm56740000gn/T/debounce1470501507670-1186788113/window-max/12.txt",
  :type :create}
```

Having played with it a bit, I think showing per watched-path is a good clean level of info for the INFO level. I'm going to push that up. Lemme know what you think!

† This actually requires a small update to also call str on the elements of the map

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: @KevinCorcoran you might want to weigh in here. I felt a little intrusive changing the basic Event format, but I felt like the code was significantly cleaner because of it. It would probably help to read the commits in order rather than the final diff to see the convolution that builds in 7886559 before being cleaned up in d850930

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +92,58 @@
...
+  within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (.pollEvents watch-key)]
+        events (get-event-maps-from-key watch-key)
+        time-limit (+ (System/currentTimeMillis) window-max)]
     (.reset watch-key)
-    [watch-key events]))
+    (watch-new-directories! events watcher)
+      (loop [events' events]
```
KevinCorcoran: There are 2 extra spaces of indentation on this line and the rest of this function.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +92,58 @@
...
-  Returns the native WatchKey and WatchEvents"
+(schema/defn retrieve-events :- [Event]
+  "Blocks until an event the watcher is concerned with has occured. Will then
+  poll for a new event, waiting at least `window-min` for a new event to
+  occur. Will continue polling for as long as there are new events that occur
+  within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (.pollEvents watch-key)]
+        events (get-event-maps-from-key watch-key)
```
KevinCorcoran: Renaming this binding to `initial-events` or similar will allow `events'` below to be renamed to just `events`.

KevinCorcoran: Also, do we need to check if this event set is empty?  (i.e. can the first call to `.pollEvents` return an empty list?)

KevinCorcoran: ... if so, we probably don't want to start looping in that case.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +92,58 @@
...
                       (and (= :create (:type event))
-                           (fs/directory? (:path event))))]
+                           (fs/directory? (:full-path event))))]
     (DirWatchUtils/registerRecursive (:watch-service watcher)
                                      (->> events
                                           (filter dir-create?)
-                                          (map #(.toPath (:path %)))))))
+                                          (map #(.toPath (:full-path %)))))))
+
+(schema/defn get-event-maps-from-key :- [Event]
```
KevinCorcoran: This name is a little unconventional, I'd probably name it `watch-key->events` or something similar.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +92,58 @@
...
-    [watch-key events]))
+    (watch-new-directories! events watcher)
+      (loop [events' events]
+        (if-let [waiting-key (.poll (:watch-service watcher) window-min window-units)]
+          (let [waiting-events (get-event-maps-from-key waiting-key)]
+            (watch-new-directories! waiting-events watcher)
+            (.reset waiting-key)
+            (if (< (System/currentTimeMillis) time-limit)
+              (recur (into events' waiting-events))
+              (into events' waiting-events)))
```
KevinCorcoran: `concat` (instead of `into`) here is more idiomatic, I think, and I'd guess more efficient too.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +92,58 @@
...
+        events (get-event-maps-from-key watch-key)
+        time-limit (+ (System/currentTimeMillis) window-max)]
     (.reset watch-key)
-    [watch-key events]))
+    (watch-new-directories! events watcher)
+      (loop [events' events]
+        (if-let [waiting-key (.poll (:watch-service watcher) window-min window-units)]
+          (let [waiting-events (get-event-maps-from-key waiting-key)]
+            (watch-new-directories! waiting-events watcher)
+            (.reset waiting-key)
```
KevinCorcoran, line 112: Is the order of the calls to `watch-new-directories!` and `.reset` important?  They are in the opposite order above.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran: Oh, I just now saw the discussion at https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/pull/8#discussion_r73751528

I still think it would be good to check if `.pollEvents` returned an empty list and, if so, not loop.

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
```diff
@@ -6,9 +6,15 @@
   "Schema for an event on a file watched by this service."
   (schema/if #(= (:type %) :unknown)
     {:type (schema/eq :unknown)
-     :path (schema/pred nil?)}
+     :count schema/Int
+     :watched-path File
+     :changed-path (schema/pred nil?)
+     :full-path (schema/pred nil?)}
```
KevinCorcoran: I just mentioned this on one of the previous PRs in this repo but I'll restate it here for convenience ...

`(schema/pred nil?)` looks very strange to me, and I think `schema/optional-key` is far more conventional.

KevinCorcoran: ... or `(schema/maybe File)`.

Also, what is `full-path`?  It's not obvious so it probably needs to be defined in the docstring.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran: Logging a map with the same values at keys of slightly different names seems really confusing to me.  I think we should just log things as they are - no need to have specific key names just for the sake of logging.  Unless I'm missing something.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -35,27 +35,32 @@
...
+  {:context changed-path
+   :count count
+   :kind type})
+
+(schema/defn format-for-consumers
+  [{:keys [full-path type]} :- Event]
+  {:path full-path
+   :type type})
+
+(schema/defn format-for-users
```
KevinCorcoran: The name of this function seems misleading since it is only used in a log statement.

Also, it doesn't seem to have much effect except logging a stringified-file instead of a `File` object.  It doesn't seem usefully-different from `format-for-consumers` - and I wonder if the logging call which uses it could just be deleted/combined with the other one at `trace` immediately after it.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -35,27 +35,32 @@
...
+
+(schema/defn format-for-debugging
+  [{:keys [changed-path count type]} :- Event]
+  {:context changed-path
+   :count count
+   :kind type})
+
+(schema/defn format-for-consumers
+  [{:keys [full-path type]} :- Event]
+  {:path full-path
```
KevinCorcoran: This seems like we're no longer conforming to the schema since `:path` isn't in the `Event` schema.

If we want to do something like this it seems like we might need to have separate internal/external schemas for events.  That might be a bit heavy-handed at the moment, so I wonder if we can simply remove this function and pass the unmodified `Event` objects into the callbacks.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -40,11 +40,11 @@
...
                        (.context event))
         full-path (when-not (= :unknown kind)
-                    (-> watched-path (.resolve changed-path) fs/file))]
+                    (-> watched-path (.resolve changed-path) (.toFile)))]
     {:type kind
      :count (.count event)
-     :watched-path (fs/file watched-path)
-     :changed-path (fs/file changed-path)
+     :watched-path (.toFile watched-path)
+     :changed-path (when changed-path (.toFile changed-path))
```
KevinCorcoran: > We cannot call `me.raynes.fs/file` on the relative path provided by
> `WatchEvent#context()` without it being resolved to an absolute path
> relative to the CWD, which is incorrect behavior.

This sentence doesn't seem to make sense to me.  By definition, absolute paths aren't relative to any directory, that's what makes them "absolute".

Is something actually broken or is this just about a desire to have these paths be relative instead of absolute?  If it's the former, shouldn't this break (all of?) the tests?  If the latter, I think I'm in support of that, but if it's that important it might be worth documenting somewhere, such as the docstring on the `Event` schema.

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
justinstoller: wrt `(schema/maybe File)`. If I understand that correctly means `nil` or `File`, which isn't the possible valid values IF we're limiting this to what is valid for `:unknown` events.

I'm in favor of saying that **if** the event type is `:unknown` the values **must** be `nil`, if  the event type is **not** `:unknown` they must be a File.
OR, we **don't differentiate** between event types and just say that the keys in question may be `nil`.

justinstoller: FWIW, prior to being `(schema/pred nil?)` they were `(schema/eq nil)` which I think is functionally the same.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller, line 112: Good question! My understanding is:

We want to call `watch-new-directories` asap so that there's as little chance as possible that we're missing nested directory creation.

We want to call `reset` before we attempt to `take` or `poll` for a `watch-key` so we can potentially get new events from it.

We don't need to call `watch-new-directories` if `events` is empty, though it shouldn't hurt.
We need to call `reset` even if `events` is empty.

I think I should switch the order above this loop.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: I had it as that for a little bit locally. But I thought that the `stuff->things` pattern was only used when translating or coercing from one type to another. In that vein I could see renaming `clojurize` to `watch-event->event`, but the meaningful portion of this function seemed to call an accessor-like function on the passed in Java object. So I went with a java-y accessor-like name. Is the `stuff->things` naming convention often used for interop like that? I don't have much experience with it.

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
KevinCorcoran: > wrt (schema/maybe File). If I understand that correctly means nil or File, 

right 

> ... which isn't the possible valid values IF we're limiting this to what is valid for :unknown events.

what else can it be?

> I'm in favor of saying that if the event type is :unknown the values must be nil, if the event type is not :unknown they must be a File.

I like this approach but I would rather we just omit the keys entirely instead of enforcing `nil` values for them.

> OR, we don't differentiate between event types and just say that the keys in question may be nil

Yeah, we could go this way but it'd be nice to not pollute the "happy path" schema since the unknown events are so different and funky.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -396,6 +396,53 @@
...
+                     (swap! actual-events concat e)
+                     (swap! actual-calls inc))
+          app (tk/boot-services-with-config watch-service-and-deps {})
+          svc (tk-app/get-service app :FilesystemWatchService)
+          event-cadence (quot watch-core/window-min 2)]
+      ;; We trigger half the number of events that could independently occur
+      ;; without exceeding the window-max, we trigger them at half the wait
+      ;; of window-min to ensure they will all be caught even if the host
+      ;; running the tests is slow. This requires that window-max be at least
+      ;; twice the duration as window-min.
```
KevinCorcoran: How is there an upper bound on the "number of events that could independently occur without exceeding the window-max"?  Ah, I see, it's because of `event-cadence` ... perhaps worth mentioning ... "number of events that could independently occur, given a fixed cadence, ..."

> we trigger them at half the wait of window-min to ensure they will all be caught even if the host running the tests is slow.

Does it actually "ensure" that or just "increase the likelihood" of that?  To say it another way: is there a race condition here?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: We create the (absolute) `:full-path` by resolving the (relative) `:changed-path` relative to the (absolute) `:watched-path`, when we call `fs/file` on the `:changed-path` it resolves it to an absolute path relative to the processes current working directory.

Here's an mock up of the before and after data:
before this change:

```
{:type :create
 :count 1
 :watched-path java.io.File["/var/folders/4l/single-path-test/"]
 :changed-path java.io.File["/Users/justin/code/tk-fs-watcher/first-file"]
 :full-path java.io.File["/var/folders/4l/single-path-test/first-file"]}
```

after this change:

```
{:type :create
 :count 1
 :watched-path java.io.File["/var/folders/4l/single-path-test/"]
 :changed-path java.io.File["first-file"]
 :full-path java.io.File["/var/folders/4l/single-path-test/first-file"]}
```

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -396,6 +396,53 @@
...
+      ;; We trigger them for what should be twice the duration of window-max
+      ;; however to see the loop properly short circuited. Depending on the
+      ;; speed of the test runner host, we could see between 2 and 4 callbacks
+      ;; without seeing failures elsewhere in the suite.
+      (testing "multiple callbacks are triggered if polling exceeds `window-max`"
+        (let [test-dir (fs/file root-dir "window-max")
+              event-nums (* (quot watch-core/window-max watch-core/window-min) 4)
+              files (map #(fs/file test-dir (str % ".txt")) (range 0 event-nums))
+              expected-events (set (map (fn [f] {:path f :type :create}) files))]
+          (is (> event-nums 2))
```
KevinCorcoran: This is an interesting assertion because it seems to imply some sort of contract about valid values of `window-max` and `window-min` ... meaning, we couldn't just change them without breaking this assertion.  I guess that's probably true of the previous `testing` block as well, but it's rather more explicit here.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -396,6 +396,53 @@
...
+          (is (fs/mkdirs test-dir))
+          (watch! svc test-dir callback)
+          (reset! actual-events [])
+          (reset! actual-calls 0)
+          (doseq [f files] (fs/touch f) (Thread/sleep event-cadence))
+          (is (= expected-events (wait-for-events actual-events expected-events)))
+          (is (>= @actual-calls 2)))))))
```
KevinCorcoran: Should you add `(is (<= @actual-calls 4))` ?

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -396,6 +396,53 @@
...
+          (is (fs/mkdirs test-dir))
+          (watch! svc test-dir callback)
+          (doseq [f files] (fs/touch f) (Thread/sleep event-cadence))
+          (is (= expected-events (wait-for-events actual-events expected-events)))
+          (is (= @actual-calls 1))))
+      ;; We trigger events at the same cadence as above to ensure that we will
+      ;; be causing events within window-min and continuing the polling loop.
+      ;; We trigger them for what should be twice the duration of window-max
+      ;; however to see the loop properly short circuited. Depending on the
+      ;; speed of the test runner host, we could see between 2 and 4 callbacks
```
KevinCorcoran, line 432: Can you explain the "between 2 and 4" part a bit more?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran: It's a common naming convention for any sort of "conversion" function.  Also, it's somewhat unconventional to name Clojure functions `get-*` - that one's far from hard-and-fast, but you don't see it all that much, and Stuart Sierra mentioned that in a blog post of his a while back.

KevinCorcoran: In general, Clojure naming conventions don't necessarily follow Java naming conventions.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
KevinCorcoran: The new format looks good.  My only thought would be to be a bit more explicit with the key names, like `:changed-relative-path` and `:changed-absolute-path`.  

On second thought ... it's actually pretty weird to return two separate `File` objects for a single path on the filesystem.  Which makes me wonder if we could just have a single `:changed-path` which is absolute.  Is there any specific need for the relative path to show up in these Events?

I still wonder if there should be a test to go along with this change, since it was clearly broken before but was not caught by the existing tests.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
justinstoller: It's not true of the code itself, but in order to write tests around it I had to make some assumptions. This could be put a different way: testing that we are debouncing events, but are allowing multiple callbacks when those events come in outside of our cutoff, requires at least more than two events to test.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
justinstoller: It increases the likelihood.

I definitely think this test has race issues it. I'm not sure the best way to test this - a quiet period and a timeout in a background thread at the integration level - without there being a race condition. My hope was to make the temporal relationships and assumptions explicit and fail the test in a more obvious way if the test runners performance goes out of the bands where this test would perform reliably.

My assumption when writing this test was that if we were seeing slowdowns large enough that something triggered at half our window-min could exceed our window-max we would either be seeing slow downs that would affect other tests (ie if what should happen in 50ms was taking over 2s we were in some kind of failed state on that host and we probably shouldn't trust our results elsewhere either) or our default settings are poorly chosen (ie we can't really test this reliably if we set our window-min to 2ms and our window-max to 8ms when we expect events to reach us within 10ms of when they occur).

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
justinstoller, line 432: Normal case there will be two calls to the registered callbacks, as the load on the system increases it will rise to 4 after that point additional load will cause the calls to quickly increase to `event-nums` since at that point events are taking more than twice as long as expected (and our expected cadence is half of window-min).

At that point the test prior will start producing multiple calls to the callbacks instead of the debounced one.

justinstoller:
Heeey @KevinCorcoran ! I think I addressed all your comments. I left a commit per comment so matching them up would be easier. Happy to squash them into a couple logical ones after we agree they're the right changes to make!

💩    (:

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
KevinCorcoran: I think race conditions are going to be inevitable in any non-trivial test cases like this.  I think this comment should just be updated to clarify the wording - specifically "ensure" is a little misleading, I think.

src/clj/puppetlabs/trapperkeeper/services/protocols/filesystem_watch_service.clj
```diff
@@ -6,9 +6,13 @@
...
     {:type (schema/eq :unknown)
-     :path (schema/pred nil?)}
+     :count schema/Int
+     :watched-path File}
     {:type (schema/enum :create :modify :delete)
-     :path File}))
+     :count schema/Int
+     :watched-path File
+     :changed-path File
+     :full-path File}))
```
KevinCorcoran: We had a comment thread going which I think got lost in [an outdated diff](https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/pull/8#discussion_r73890035).

To summarize:

It's pretty weird to return two separate `File` objects for a single path on the filesystem.  That makes me wonder if we could just have a single `:changed-path` which is absolute.  Is there any specific need for the relative path to show up in these `Event` objects?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -81,42 +83,55 @@
...
-  Returns the native WatchKey and WatchEvents"
+(schema/defn retrieve-events :- [Event]
+  "Blocks until an event the watcher is concerned with has occured. Will then
+  poll for a new event, waiting at least `window-min` for a new event to
+  occur. Will continue polling for as long as there are new events that occur
+  within `window-min`, or the `window-max` time limit has been exceeded."
   [watcher :- (schema/protocol Watcher)]
   (let [watch-key (.take (:watch-service watcher))
-        events (.pollEvents watch-key)]
+        events (watch-key->events watch-key)
```
KevinCorcoran: I left this comment before but I think it got buried too ...

Renaming this binding to `initial-events` or similar will allow `events'` below to be renamed to just `events`.

KevinCorcoran:
🙈  A few final nitpicks.  The only issue of significance I see with this PR in its current state is the issue with including two `File` objects for both the relative and absolute paths in the `Event` schema.

KevinCorcoran:
LGTM! 👍 just needs some squashin'

justinstoller:
Squashed!

@KevinCorcoran I created a test for the happy path within `clojurize` to ensure our paths were correct per https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/pull/8#discussion_r73912442. But after writing it I realized that when we combined `:changed-path` and `:full-path` we began testing the path resolution we were concerned about with every integration test. So the test ended up not being very helpful and after reviewing with @haus we figured we should just can it. It still lives [here](https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/commit/d187cb26ee210cdd9f1b0666f458d31f9641b19e) if you'd like to peruse it.

PR #9, opened 2016-08-09
(PE-16860) Ship TK-FS-Watcher 1.0

justinstoller:


justinstoller:
I believe the editing of the project.clj was the correct thing to do when bumping after the dev work has taken place. Lemme know if not!

PR #10, opened 2016-08-09
(maint) Return the context we have been passed

justinstoller:


haus:
I dunno. This seems risky.

justinstoller:
Maybe it's unnecessary? Nothing in any of tests for this repo **or** its consumer breaks because of this....

haus:
See. You should never change anything.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service.clj
```diff
@@ -12,7 +12,7 @@
 
   (init
     [this context]
-    {:watchers (atom [])})
+    (assoc context :watchers (atom [])))
```
KevinCorcoran, line 15: Was this change strictly necessary or did it just seem like a good idea?

Isn't `(= context {})` at this point?

justinstoller, line 15: It wasn't failing a test... but it seemed that unless we could guarantee tk-fs-watcher was always initialized first we would at some point swallow someone's context.

FWIW, w came across this because we did have a test that produced error messages when we initially promoted tk-fs-watcher 1.0 into pe-file-sync. It was ensuring that the storage service shutdown during its init phase if passed an invalid config. After a while we realized that with tk-fs-watcher no longer having a dependency on tk-scheduler the storage-service no longer needed it either, and the test was failing because  the scheduler wasn't getting to its init phase during the test, but was running its stop phase, failing to deref an atom it expected to exist in its context that it would have put in there during its init. That led us to search for things-that-could-be-trashing-a-services-context-during-init and we found this.

KevinCorcoran, line 15: (x-posting from chat for posterity ...)

`context` here is the service-specific context, not the global TK context for all services.  Thus, for any service, it is always `{}` at the beginning of the `init` function.  So, this change has no effect.

PR #12, opened 2016-08-11
(TK-389) Debouncing test polish

KevinCorcoran:
👋 @justinstoller 

In the first commit, there's some minor code cleanup and a few TODOs in which I have some questions I'm hoping you can answer, after which I'll touch-up that commit.

In the second commit, I just deleted the last assertion, which was added at my suggestion IIRC, and which recently failed in CI.

### DIFF
--- a/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
@@ -432,9 +432,8 @@
       ;; We trigger events at the same cadence as above to ensure that we will
       ;; be causing events within window-min and continuing the polling loop.
       ;; We trigger them for what should be twice the duration of window-max
-      ;; however to see the loop properly short circuited. Depending on the
-      ;; speed of the test runner host, we could see between 2 and 4 callbacks
-      ;; without seeing failures elsewhere in the suite.
+      ;; however to see the loop properly short circuited. The callback should be
+      ;; invoked at least twice.
       (testing "multiple callbacks are triggered if polling exceeds `window-max`"
         (let [test-dir (fs/file root-dir "window-max")
               event-nums (* (quot watch-core/window-max watch-core/window-min) 4)
@@ -447,9 +446,7 @@
           (reset! callback-invocations 0)
           (doseq [f files] (fs/touch f) (Thread/sleep event-cadence))
           (is (= expected-events (wait-for-events actual-events expected-events)))
-          (is (>= @callback-invocations 2))
-          (is (<= @callback-invocations 4)
-              "System speed is slower than expected to reliably run this test"))))))
+          (is (>= @callback-invocations 2)))))))
 
 ;; Here we create a stub object that implements the WatchEvent interface as
 ;; the concrete class is a private inner class. See:
--- a/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
+++ b/test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
@@ -398,33 +398,37 @@
             :error))))))
 
 (deftest ^:integration debouncing-test
-  (testing "Default debouncing behavior"
+  (testing "debouncing"
     (let [root-dir (fs/temp-dir "debounce")
           actual-events (atom [])
-          actual-calls (atom 0)
+          callback-invocations (atom 0)
           callback (fn [e]
                      (swap! actual-events concat e)
-                     (swap! actual-calls inc))
+                     (swap! callback-invocations inc))
           app (tk/boot-services-with-config watch-service-and-deps {})
           svc (tk-app/get-service app :FilesystemWatchService)
           event-cadence (quot watch-core/window-min 2)]
-      ;; We trigger half the number of events that could independently occur
-      ;; without exceeding the window-max, given the fixed event-cadence
+      ;; We trigger half the minimum number of events that could independently
+      ;; occur without exceeding the window-max, given the fixed event-cadence
       ;; set above, we set the event-cadence to half of window-min to increase
-      ;; the likelihood that they will be caught regardless of test runner
-      ;; speed. This requires that window-max be at least four times the
+      ;; the likelihood that they will debounced into a single callback
+      ;; invocation. This requires that window-max be at least four times the
       ;; duration of window-min.
-      (testing "multiple events trigger one callback when within `window-min` and `window-max`"
+      (testing "multiple events trigger one callback when within window-min and window-max"
         (let [test-dir (fs/file root-dir "window-min")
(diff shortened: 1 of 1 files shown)


KevinCorcoran:
The Travis CI failures are both in `multiple-watch-dirs-test`, which is unaffected by this PR.  And clearly caused by our old friend, the JDK bug.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -398,39 +398,50 @@
...
-          event-cadence (quot watch-core/window-min 2)]
+          event-cadence (/ watch-core/window-min 2)]
       ;; We trigger half the number of events that could independently occur
       ;; without exceeding the window-max, given the fixed event-cadence
       ;; set above, we set the event-cadence to half of window-min to increase
       ;; the likelihood that they will be caught regardless of test runner
-      ;; speed. This requires that window-max be at least four times the
+      ;; speed.
+      ;;
+      ;; TODO what does "be caught" mean?
```
justinstoller: "be debounced"?

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -398,39 +398,50 @@
...
       ;; duration of window-min.
-      (testing "multiple events trigger one callback when within `window-min` and `window-max`"
+      (testing "multiple events trigger one callback when within window-min and window-max"
         (let [test-dir (fs/file root-dir "window-min")
-              event-nums (quot (quot watch-core/window-max watch-core/window-min) 2)
-              files (map #(fs/file test-dir (str % ".txt")) (range 0 event-nums))
-              expected-events (set (map (fn [f] {:changed-path f :type :create}) files))]
-          (is (> event-nums 1))
+              ;; TODO shouldn't num-events be (quot (quot window-max event-cadence) 2) ?
+              ;; (based on the the previous comment)
```
justinstoller: It would be more likely for us to hit window-max if we used that number. But we still shouldn't hit it under normal circumstances, if you'd prefer it to be calculated like that.

The above number is a direct translation of:

> We trigger half the number of events that could independently occur without exceeding the window-max

We could in theory have one event per window-min that would trigger the code to loop again, so the minimum number of events that could trigger the window-max timeout would be `(quot window-max window-min)`.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -398,39 +398,50 @@
...
-          actual-calls (atom 0)
+          callback-invocations (atom 0)
           callback (fn [e]
                      (swap! actual-events concat e)
-                     (swap! actual-calls inc))
+                     (swap! callback-invocations inc))
           app (tk/boot-services-with-config watch-service-and-deps {})
           svc (tk-app/get-service app :FilesystemWatchService)
-          event-cadence (quot watch-core/window-min 2)]
+          event-cadence (/ watch-core/window-min 2)]
```
justinstoller: If window-min isn't cleanly divisible by 2 this will result in a rational number. This will be passed into `Thread/sleep` below. Is there a reason we would want to wait for a fraction of a millisecond?

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -398,39 +398,50 @@
...
+              ;; TODO shouldn't num-events be (quot (quot window-max event-cadence) 2) ?
+              ;; (based on the the previous comment)
+              num-events (quot (quot watch-core/window-max watch-core/window-min) 2)
+              files (for [n (range num-events)]
+                      (fs/file test-dir (str n ".txt")))
+              expected-events (set (for [f files] {:changed-path f
+                                                   :type :create}))]
+          ;; TODO this assertion strikes me as weird - it's the result of some
+          ;; division of constants - we should be able to say exactly what it is.
+          (is (> num-events 1))
```
justinstoller, line 424: Not sure what you mean? If window-min is equal to or larger than 1/4 of window-max the num-events will be one or zero. Testing debouncing doesn't work well without multiple events to debounce.

I think there's some confusion because these tests are written to not care specifically about the window-min and window-max values. Using a cadence and numbers of events that are reasonable, I think, to assume, and failing the test if we change window-min and window-max in ways that mean those assumptions are no longer valid. However, that means we're computing a lot of values as the test runs and making assertions that our assumptions hold.

We could probably just hard code values for cadence and num-events that are reasonable for the known values of window-min and window-max and having a simple assertion that the values of window-min and window-max are what they currently are set to (with a note that if they change the values in this test might need to change to still be valid).

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
KevinCorcoran: 😆 no, I doubt there would be.  The reason for this change was simply that `quot` "looked weird" and `/` is the usual way of doing division.

Actually, I think things would go 💥  in that case, since I think `Thread/sleep` requires an integer.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
justinstoller: Ahhh...

The full quote is:

> We trigger half the number of events that could independently occur without exceeding the window-max, given the fixed event-cadence set above,

Would it make more sense if it was:

> We trigger half the number of events that could independently occur without exceeding the window-max, given a fixed event-cadence,

KevinCorcoran: Hmm.  In my mind a direct translation of 

> the number of events that could independently occur without exceeding the window-max

would be 

> (quot window-max event-cadence)

Oh ... I was thinking of _maximum_ number of events.  Yeah, that's different.  💡

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
KevinCorcoran, line 424: I think it's a reasonable approach, but, yeah, it does make the test a bit harder to understand.  I think I'll just delete this TODO.

KevinCorcoran:
feedback addressed - force-pushed a new commit which removes the added TODOs and touches-up the comments as per discussion

justinstoller:
👍

KevinCorcoran:
I reverted the change to `/`

KevinCorcoran:
latest Travis CI failure is also due to the JDK 🐛

PR #13, opened 2017-02-01
(TK-430) Migrate to clj-parent 0.3.3 and bump i18n to 0.6.0

camlow325:
This PR migrates tk-filesystem-watcher to using clj-parent version 0.3.3
for its dependencies.  This commit also adds i18n as a plugin dependency, to
allow for message POT files to be regenerated as needed in CI.  This
commit also upgrades the lein version to 2.7.1 in the .travis.yml
file since 2.7.1 is the minimum version required for compatibility
with lein/clj-parent.

This PR also removes tk-scheduler as a dependency since none of the
production code was referencing it and the tests that were referencing it
did not seem to be needing to do so.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,24 +1,26 @@
-(def tk-version "1.4.1")
-(def ks-version "1.3.1")
-
 (defproject puppetlabs/trapperkeeper-filesystem-watcher "1.0.2-SNAPSHOT"
   :description "Trapperkeeper filesystem watcher service"
   :url "https://github.com/puppetlabs/trapperkeeper-filesystem-watcher"
   :license {:name "Apache License, Version 2.0"
             :url "http://www.apache.org/licenses/LICENSE-2.0"}
 
+  :min-lein-version "2.7.1"
+
+  :parent-project  {:coords [puppetlabs/clj-parent "0.3.3"]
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -28,12 +29,15 @@
   :source-paths ["src/clj"]
   :java-source-paths ["src/java"]
 
-  :profiles {:dev {:dependencies [[puppetlabs/trapperkeeper ~tk-version
+  :profiles {:dev {:dependencies [[puppetlabs/trapperkeeper nil
                                    :classifier "test"
                                    :scope "test"]
-                                  [puppetlabs/kitchensink ~ks-version
+                                  [puppetlabs/kitchensink nil
```
jpinsonault: Does putting `nil` here mean it'll pull the version from the parent project? Sucks that we have to leave a `nil` lying around

camlow325: Yeah, the `nil` would signify that that the version is pulled in from the parent project.  It may actually work without the `nil` in later lein-parent versions, can't remember.  I can try removing it and see if it still works.

camlow325: Yeah, turns out the `nil` references don't appear to be needed after all.  I removed those in 45414655.

PR #14, opened 2017-02-04
(MAINT) Add i18n scaffolding

camlow325:
This commit adds basic i18n files - Makefiles and an initial
messages.pot file.


rlinehan (approved):
:+1:

PR #17, opened 2017-08-04
(TK-448) Add support for non-recursive directory watching

MosesMendoza:
* For the watcher protocol implementation, add a new field `recursive`. The field is initially populated as an atom with a nil value. Upon the first watch directory added, the field is set "permanently" to the value specified for the  :recursive option. We do so by enforcing that upon addition of any new directories to watch, the :recursive option must be the same as previously supplied, or we throw an exception. Also add a docstring to watch-new-directories! to clarify its behavior.

* Plumb recursive option through service. Update the required methods in the watch service to take the recursive setting into acount when deciding whether or not to recurse into directories that are watched and add them to the list of watched directories. With this, if a user specificies :recursive false, we limit watching to only the directory specified by `add-watch-dir!` and no others.

* Add new test utilities that allow us to check if a specific event - and only that event - have been registered by the watch service. We do this by immediately doing our comparison check on the first event that is registered.

  We are in the odd position of needing to test that something -doesn't happen-, not that it does. We can't just test that no events are registered, because that might happen for a number of reasons (ie the underlying watcher service failed). So instead we have a canary event alongside every operation that should not generate an event, and then we check that the only event that is registered is the canary event.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -89,12 +89,12 @@
 (defn create-watcher
   ([]
    (create-watcher {:recursive true}))
-  ([options]
+  ([{:keys [recursive] :as options}]
    (validate-watch-options! options)
    (map->WatcherImpl
      {:watch-service (.newWatchService (FileSystems/getDefault))
       :callbacks (atom [])
-      :recursive (atom (:recursive options))})))
+      :recursive (atom recursive)})))
 
 (schema/defn watch-new-directories!
   "Given an initial set of events and a watcher, identify any events that
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -54,10 +54,10 @@
 
 (defn validate-watch-options!
   [options]
-  (when-not (= true (:recursive options))
+  (when-not (instance? Boolean (:recursive options))
     (throw
       (IllegalArgumentException.
-        (trs "Support for non-recursive directory watching not yet implemented")))))
+        (trs "Must pass a boolean value for :recursive (directory watching) option")))))
 
 (defrecord WatcherImpl
   [watch-service callbacks]
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -4,7 +4,7 @@
             [schema.core :as schema]
             [puppetlabs.i18n.core :refer [trs]]
             [puppetlabs.kitchensink.core :as ks]
-            [puppetlabs.trapperkeeper.services.protocols.filesystem-watch-service :refer [Event Watcher]])
+            [puppetlabs.trapperkeeper.services.protocols.filesystem-watch-service :refer [Event Watcher] :as watch-protocol])
   (:import (clojure.lang Atom IFn)
            (java.nio.file StandardWatchEventKinds Path WatchEvent WatchKey FileSystems ClosedWatchServiceException)
            (com.puppetlabs DirWatchUtils)))
@@ -53,50 +53,48 @@
        :changed-path (.. watched-path (resolve (.context event)) (toFile))})))
 
 (defn validate-watch-options!
+  "Validate that the options supplied include a valid Boolean value for key :recursive."
   [options]
   (when-not (instance? Boolean (:recursive options))
     (throw
       (IllegalArgumentException.
         (trs "Must pass a boolean value for :recursive (directory watching) option")))))
 
-(schema/defn set-final-recursive!
-  "Enforce only setting the recursive option for a watcher once, and fail on
-  attempted modification, given an atom representing the value of the recursive
-  setting for a watcher and a new boolean value."
-  [recursive :- Atom
-   recursive-option :- Boolean]
-  (if (nil? @recursive)
-    (reset! recursive recursive-option)
-    (when-not (= @recursive recursive-option)
-      (throw
-        (IllegalArgumentException.
-          (trs "Watcher already set to :recursive {0}, cannot change to :recursive {1}"
-            @recursive recursive-option))))))
-
 (defrecord WatcherImpl
   [watch-service callbacks recursive]
   Watcher
   (add-watch-dir!
-    [this dir options]
-    (validate-watch-options! options)
-    ;; the first directory we watch defines the recursive behavior of this watcher - once set it cannot be changed
-    (set-final-recursive! recursive (:recursive options))
-    (let [recursive (:recursive options)
-          watched-path (.toPath (fs/file dir))]
-      (if recursive
+    [this dir]
+    (let [watched-path (.toPath (fs/file dir))]
+      (if @recursive
         (DirWatchUtils/registerRecursive watch-service [watched-path])
         (DirWatchUtils/register watch-service watched-path))))
+
+  (add-watch-dir!
+    [this dir options]
+    (validate-watch-options! options)
+    ;; The value of recursive was already set by `create-watcher`, possibly by specifying an option.
+    ;; We validate that the option supplied to `add-watch-dir!` has the same value.
+    (when-not (= @recursive (:recursive options))
+      (throw
+        (IllegalArgumentException.
+          (trs "Watcher already set to :recursive {0}, cannot change to :recursive {1}"
+            @recursive (:recursive options)))))
+    (watch-protocol/add-watch-dir! this dir))
 
   (add-callback!
     [this callback]
     (swap! callbacks conj callback)))
 
 (defn create-watcher
-  []
-  (map->WatcherImpl
-    {:watch-service (.newWatchService (FileSystems/getDefault))
-     :callbacks (atom [])
-     :recursive (atom nil)}))
+  ([]
+   (create-watcher {:recursive true}))
+  ([options]
+    (validate-watch-options! options)
+    (map->WatcherImpl
+      {:watch-service (.newWatchService (FileSystems/getDefault))
+      :callbacks (atom [])
+      :recursive (atom (:recursive options))})))
 
 (schema/defn watch-new-directories!
   "Given an initial set of events and a watcher, identify any events that
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -79,7 +79,7 @@
       (throw
         (IllegalArgumentException.
           (trs "Watcher already set to :recursive {0}, cannot change to :recursive {1}"
-            @recursive (:recursive options)))))
+               @recursive (:recursive options)))))
     (watch-protocol/add-watch-dir! this dir))
 
   (add-callback!
@@ -90,9 +90,9 @@
   ([]
    (create-watcher {:recursive true}))
   ([options]
-    (validate-watch-options! options)
-    (map->WatcherImpl
-      {:watch-service (.newWatchService (FileSystems/getDefault))
+   (validate-watch-options! options)
+   (map->WatcherImpl
+     {:watch-service (.newWatchService (FileSystems/getDefault))
       :callbacks (atom [])
       :recursive (atom (:recursive options))})))
 
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -81,7 +81,8 @@
     (validate-watch-options! options)
     ;; the first directory we watch defines the recursive behavior of this watcher - once set it cannot be changed
     (set-final-recursive! recursive (:recursive options))
-    (let [watched-path (.toPath (fs/file dir))]
+    (let [recursive (:recursive options)
+          watched-path (.toPath (fs/file dir))]
       (if recursive
         (DirWatchUtils/registerRecursive watch-service [watched-path])
         (DirWatchUtils/register watch-service watched-path))))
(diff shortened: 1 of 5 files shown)


MosesMendoza:
Will squash commits on review/approval

MosesMendoza:
The test failing in travis passes for me locally... it doesn't appear that it _should_ be failing, so i'm wondering if there's some sort of timing thing going on

MosesMendoza:
Yep, now some unrelated test appears to be failing, and meanwhile:
```
☂ : lein -U test :all

Ran 10 tests containing 71 assertions.
0 failures, 0 errors.
```

MosesMendoza:
~The test seems to consistently fail in travis so i'm going to close and try to repro that failure~ I haven't been able to reproduce locally, but I did re-run travis against the current master branch commit, c18b025, and the exact same test failed.

It appears to be registering both a create and modify entry for a file getting created in the `nested watch directories` test. Some conversation in TK-387 leads me to think this is a known issue?

from @justinstoller: 
> On Linux for every created file there is at least a CREATE and MODIFY event that occurs.

Though it wasn't clear what the outcome was.

MosesMendoza:
Oh now it passes.

justinstoller:
Had a conversation with Moses about this because I was somewhat unsure why we chose this route. Lemme see if I can put it down here and see if that's something folks think is fine going forward.

The current interface has a protocol function `add-watch-dir!` that takes an options map that is always expected to have the key `:recursive` with the value `true`. This implies that there will be a way to add individual directories (paths) to a watcher that could be recursive or not.

Now that we're implementing non-recursive watching we don't want to watch per path but per watcher, and the perceived cost of implementing per path recursive watching means we feel like it is prohibitively expensive to implement.

It seems like setting recursion per-watcher would be a natural fit for adding or modifying to the service protocol, `create-watcher` could default to calling `create-recursive-watcher` while the new consumer could call a `create-non-recursive-watcher` for example.

But, for backwards compatibility, we don't want to change the signature of `add-watch-dir!` and so we've created a config value that the watcher holds in an atom. That atom is set as a side effect the first time that `add-watch-dir!` is called so we don't have to put that config value in the protocol.

justinstoller:
So, it seems to me that an option on the `add-watch-dir!` function that sets per-watcher state should be moved to the interface that returns the watcher (despite being clojure this is still very Java-objecty, so I would think of the `create-watcher` protocol as the constructor). I don't think it makes *less* sense to move that option to a "constructor" or "setter" and have the current option to `add-watch-dir!` actually be optional (and deprecated). I think it should also be feasible to do this is a backwards compatible way.

However, I don't know if others agree with me on that point or if I'm missing something, if that's a bunch of work, or if we have the timeframe to do it in.

haus:
I prefer putting it in the constructor if that is doable in a non-gross backward compatible manner.

MosesMendoza:
I agree with the sentiment that the recursive option should be on the `create-watcher` "constructor". It only made sense to be on the `add-watch-dir!` function if you could actually specify the option on a per-watched directory basis.

That said, I think we're stuck between a rock and hard place here. I'm having a hard time coming up with something that accomplishes that with a good user experience that's backwards compatible.

We could add the recursive option to the constructor where it ideally would be, but what should happen when the recursive option is also passed in via `add-watch-dir!`?

I think its kind of messy. In order to be backwards compatible, the option passed to the function has to be authoritative. So if I've specified `:recursive true` to the constructor, and `:recursive false` to the `add-watch-dir!` function, we would disable recursive watching...which I think would be  surprising. Add a bit of extra complexity because we only override it if we somehow know that the option was originally specified as part of the constructor and not by a different call to `add-watch-dir!`, in which case we fail. We could say that since you've only ever been able to pass "false" as the recursive option, that's still the case - if you want "true" you have to specify it via the constructor. But then we have this vestigial options map alongside calls to `add-watch-dir!` floating around with `:recursive false` that doesn't do anything and is just ignored, which is also surprising UX.

I think the correct solution is to just stop supporting the recursive option on `add-watch-dir!` altogether, and only support it on the constructor function. It just doesn't make sense on that function. That's clearly a hard break we would do in a major release. In anticipate, we could add a warning or log entry on `add-watch-dir!` that the recursive option will be moving in an upcoming major release? Other than that though I'm hesitant to go further, and add more complexity/confusion with some mixture of options supported on both with a precedence behavior.

justinstoller:
Just to re-iterate, I don't what the perfect to be the enemy of the good, and if changing up the interface is too much work we can short circuit this convo. Having said that I disagree with...

> I think its kind of messy. In order to be backwards compatible, the option passed to the function has to be authoritative.

So, in current patch only the *first* call to `add-watch-dir!` is authoritative[1]. Subsequent calls require an option to be passed, but that option *must agree* with whatever option was passed to the first call of `add-watch-dir!`.  My thinking is that may be a more surprising behavior than if we could move that option to the "constructor" protocol function (but make it optional so older clients will continue to get a recursive watcher by default), and make the map to `add-watch-dir!` optional, but if it is supplied enforce that it must match the kind of watcher returned.


1 -  granted in our only consumer thus far we only call `add-watch-dir!` once, but I believe the interface is meant to support multiple invocations

justinstoller (approved):

MosesMendoza:
Chatted with @justinstoller - wasn't sure what the implications of:
> make the map to add-watch-dir! optional, but if it is supplied enforce that it must match the kind of watcher returned

That means we'll enforce effectively that a user can only specify a non-recursive watcher (value of `false`) via use of an option on the `create-watcher` function, the value of which will default to `true` (recursive). Then if the option is _also_ specified on `add-watch-dir!`, it also has to have a value of `false`. This isn't a backwards-incompatible change because users can upgrade cleanly - they won't have to change any option values of existing uses. And per @justinstoller's note the end result is likely an equivalent or lesser amount of surprising behavior.

I'm going to update the PR per this feedback.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -53,31 +53,53 @@
...
   (add-watch-dir!
+    [this dir]
+    (let [watched-path (.toPath (fs/file dir))]
+      (if @recursive
+        (DirWatchUtils/registerRecursive watch-service [watched-path])
+        (DirWatchUtils/register watch-service watched-path))))
+
+  (add-watch-dir!
     [this dir options]
     (validate-watch-options! options)
```
MosesMendoza, line 75: should this have a deprecation warning emitted, ie `log/debug "passing options ..."`?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -53,31 +53,53 @@
...
+  [watch-service callbacks recursive]
   Watcher
   (add-watch-dir!
+    [this dir]
+    (let [watched-path (.toPath (fs/file dir))]
+      (if @recursive
+        (DirWatchUtils/registerRecursive watch-service [watched-path])
+        (DirWatchUtils/register watch-service watched-path))))
+
+  (add-watch-dir!
```
MosesMendoza, line 73: Looked odd to me at first but apparently multi-arity protocol function definitions are achieved via separate discrete definitions.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -53,31 +53,53 @@
...
     [this callback]
     (swap! callbacks conj callback)))
 
 (defn create-watcher
-  []
-  (map->WatcherImpl
-    {:watch-service (.newWatchService (FileSystems/getDefault))
-     :callbacks (atom [])}))
+  ([]
+   (create-watcher {:recursive true}))
```
MosesMendoza, line 93: What's the standard for spacing/indentation in clojure?

justinstoller: Two spaces unless it's a list with a single item on the first line, then one.

We follow this https://github.com/bbatsov/clojure-style-guide
relatively closely, with three noticeable exceptions being we don't usually make anything private, prefer schema to pre/post assertions, and always use lists of `<package><space><one or more classes>` in imports (rather than a vectors of `<package><dot><class>`).

justinstoller (approved):

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
justinstoller: yeah

justinstoller:
I think this looks good, though I haven't built these changes out into a server build (with the other complementary PRs) to see if it solves the problem it was intended to solve.

It'd be great to get a review/merge from @haus or someone else who is, or wants to be, familar with the code base.

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
```diff
@@ -43,6 +49,25 @@
               (println "timed-out waiting for events")
               @dest)))))))
 
+(defn wait-for-exactly-event
+  "Waits for event to land in dest. On first registered event or events in dest,
+  determines whether the registered event(s) match the supplied expected
+  event. If so, returns the expected event. Otherwise returns contents of dest."
+  [dest expected-event]
+  (let [start-time (System/currentTimeMillis)]
+    (loop []
```
haus: It looks like this could be a `while` instead of a `loop`/`recur` as there are no bindings for the loop.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -53,31 +53,53 @@
...
+  ([options]
```
haus: You might consider using destructuring here in the fn args.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
MosesMendoza: Do you know if we will always automatically emit that deprecation because of our use for HA? @haus  can we update HA to use the new form?

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
MosesMendoza: ok thanks

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
MosesMendoza: why?

haus: It makes it clear which options the function is caring about.

It would matter more if there were a bunch of keys being pulled out of the map, in this case either way is fine.

MosesMendoza: ah sure. I wasn't sure but that seems reasonable

MosesMendoza: @haus how do I destructure this in order to retain both a reference to `options` since we use it as well as the value of `:recursive`?

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
MosesMendoza: So I could do this but `elapsed-time` of the recur condition `(< elapsed-time wait-time)` is dynamically redefined/rebound every iteration which appears harder with `while` and we're also waiting for the atom value to be populated.
I could do: `(while (< (- (System/currentTimeMillis) start-time) wait-time)` and avoiding binding `elapsed-time` but that seems less readable to me. It basically ends up being something like "just wait until one of these conditions is true", ie:
```
  (let [start-time (System/currentTimeMillis)]
    (while (and (empty? @dest)
                (< (- (System/currentTimeMillis) start-time) wait-time)))
    (if (empty? @dest)
      (do
        (println "timed-out waiting for events")
         @dest)
      (if (exactly-matches-event? dest expected-event)
        expected-event
        @dest)
```
this seems really weird to me so i'm sure there's a better way

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
MosesMendoza: per comment from @justinstoller this is  `{:keys [recursive] :as options}`

MosesMendoza: updated

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
MosesMendoza: chatted with @haus in meatspace, need to use binding in loop

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
MosesMendoza: debug message added

test/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service_test.clj
MosesMendoza: updated

haus:
@KevinCorcoran 💩

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -53,31 +53,55 @@
...
-    {:watch-service (.newWatchService (FileSystems/getDefault))
-     :callbacks (atom [])}))
+  ([]
+   (create-watcher {:recursive true}))
+  ([{:keys [recursive] :as options}]
+   (validate-watch-options! options)
+   (map->WatcherImpl
+     {:watch-service (.newWatchService (FileSystems/getDefault))
+      :callbacks (atom [])
+      :recursive (atom recursive)})))
```
haus, line 99: Doesn't matter too much, but this doesn't need to be an `atom` does it? It will never change.

haus:
oh travis and the infinite sadness.

haus:
I had one question but it doesn't need to block merging this.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
MosesMendoza: Yeah I think the atom bit is leftover from when we were first spiking this out

MosesMendoza:
yeah thats the same test that passes locally but that intermittently fails or passes depending on something in travis

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
haus: right

PR #21, opened 2019-10-29
(PE-27497) Continue traversing directory when a file disappears

Magisus:
This commit updates our FileVisitor to continue iterating when it
encounters a file that has disappeared. If we have a watcher on a
directory with files managed by Puppet, there's a chance that we
might iterate over it while the file is being replaced, which
creates and then destroys a lockfile. Previously, this would cause the
Visitor to throw an exception, so this commit squashes that exception.

### DIFF
--- a/src/java/com/puppetlabs/DirWatchUtils.java
+++ b/src/java/com/puppetlabs/DirWatchUtils.java
@@ -49,6 +49,18 @@ public class DirWatchUtils {
                     register(watcher, dir);
                     return FileVisitResult.CONTINUE;
                 }
+
+                @Override
+                public FileVisitResult visitFileFailed(Path file, IOException exc)
+                        throws IOException {
+                     if (exc instanceof java.nio.file.NoSuchFileException) {
+                         // The file may have disappeared since we started traversing,
+                         // (e.g. a lockfile getting cleaned up), ignore it and continue.
+                         return FileVisitResult.CONTINUE;
+                     } else {
+                         throw exc;
+                     }
+                }
             });
         }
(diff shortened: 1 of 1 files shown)


Magisus:
I'd appreciate thoughts on if there's a good way to test this... It's designed to catch a race condition that's very hard to reproduce.

src/java/com/puppetlabs/DirWatchUtils.java
```diff
@@ -49,6 +49,18 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
                     register(watcher, dir);
                     return FileVisitResult.CONTINUE;
                 }
+
+                @Override
+                public FileVisitResult visitFileFailed(Path file, IOException exc)
+                        throws IOException {
+                     if (exc instanceof java.nio.file.NoSuchFileException) {
```
joshcooper, line 56: There's also a `NotDirectoryException` though I think that would be bizarre as it would mean Java tried to list the directory's children, but the directory was converted to a file or symlink. So 👍 as is.

joshcooper (approved):

justinstoller:
I think of a good way to test this w/o creating custom test visitors.

justinstoller (approved):

Magisus:
Yeah that's the only thing I could think of too, some way to make them iterate at a slower, more predictable rate. Seems pretty harmless, might just be best verified by seeing if it makes the PE transients go away.

jpartlow (approved):

PR #23, opened 2019-10-30
(maint) Add SNAPSHOT to version

Magisus:


project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject puppetlabs/trapperkeeper-filesystem-watcher "1.1.0"
+(defproject puppetlabs/trapperkeeper-filesystem-watcher "1.1.0-SNAPSHOT"
```
justinstoller: I think you want "1.1.1-SNAPSHOT"

Magisus: Oh yeah, good catch.

PR #24, opened 2020-02-03
(PE-27497) Rescue NoSuchFileException when registering

justinstoller:
Some implementations of temporary "files" creates a directory instead of
a true file (mkdir is apparently required on older windows systems for
atomic file creation).  This triggers us to register the directory for
watching when watching the parent directory recursively. This behavior
creates a race condition where the "lockfile" may be deleted while its
registration is pending.

Checking for the file/dirctory to exist within our implementation of
register doesn't really buy us anything since we're called immediately
during the "vistDirectory" action of the FileTreeWalker - ie. it
existed when the previous function was called and the previous function
just defers to us. (This theory holds up during testing)

### DIFF
--- a/src/java/com/puppetlabs/DirWatchUtils.java
+++ b/src/java/com/puppetlabs/DirWatchUtils.java
@@ -26,11 +26,17 @@ public class DirWatchUtils {
      */
     public static void register(final WatchService watcher,
                                 final Path dir) throws IOException {
-        dir.register(watcher,  new WatchEvent.Kind[]{
-                StandardWatchEventKinds.ENTRY_MODIFY,
-                StandardWatchEventKinds.ENTRY_CREATE,
-                StandardWatchEventKinds.ENTRY_DELETE},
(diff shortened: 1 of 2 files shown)


justinstoller:
I need to spin up a PE instance and re-test this - I changed from rescuing `IOException` to rescuing `FileNotFoundException` since the PE reproducer went away.

Not sure if I should add a test? I didn't originally since trying to write a race-y test is pretty difficult / I don't think we usually get much bang for buck there. However I could just write a test that called `DirWatchUtils.register` with a non-existent file to ensure the exception raised doesn't change on us??

justinstoller:
@puppetlabs/puppetserver-maintainers

Magisus:
I think I do kind of like the idea of the second test you outline. But this looks good! Thanks for tracking it down.

mwaggett (approved):

justinstoller:
Good call on the test   ( ;

mwaggett (approved):

Magisus (approved):

PR #29, opened 2024-05-14
(maint) update clj-parent to latest, avoid expensive logging

jonathannewman:
This updates clj-parent to 4.20 and removes an exclusion for clojure.

It also conditionalizes some potentially expensive logging when the logging levels aren't used. Specifically it avoids a `doseq` and `count` of the events (which can be thousands as seen in the field) when debug isn't enabled, and avoids a potentially expensive `print-to-string` when trace isn't enabled.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -6,12 +6,10 @@
 
   :min-lein-version "2.9.1"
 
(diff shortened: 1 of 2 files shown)


steveax (approved):
👍

PR #30, opened 2024-05-15
(maint) logging level optimization, streamline shutdown

jonathannewman:
See individual commit comments.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -145,23 +145,24 @@
           events))
       initial-events)))
 
-
 (schema/defn process-events!
-  "Process for side-effects any events that occured for watcher's watch-key"
+  "Process for side effects any events that occurred for watcher's watch-key"
   [watcher :- (schema/protocol Watcher)
    events :- [Event]]
-  (let [callbacks @(:callbacks watcher)
-        events-by-dir (group-by :watched-path events)]
+  (let [callbacks @(:callbacks watcher)]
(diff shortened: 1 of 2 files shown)


src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_service.clj
```diff
@@ -2,27 +2,44 @@
...
+
+    (doseq [watchers-future @(:watchers-futures context)]
+      (try
+        (future-cancel watchers-future)
+        (loop [tries 100]
+          (if (and (pos? tries) (not (future-done? watchers-future)))
+            (do
+              (Thread/sleep 10)
+              (recur (dec tries)))
+            (log/debug (trs "Future completed with {0} tries left" tries))))
```
justinstoller: I think it would be better to say "Future completed after X tries" rather than asking folks know that it will try 100 times.

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -145,39 +145,47 @@
...
     ;; avoid doing a potentially expensive print-to-string when we aren't logging at :trace
     (when (log/enabled? :trace)
       (log/tracef "%s\n%s"
                   (trs "Events:")
                   (ks/pprint-to-string events)))
+
     (doseq [callback callbacks]
       (callback events))))
 
 (schema/defn watch!
```
justinstoller: we should add that this returns a Future now that we depend on that behavior

jonathannewman: Good idea.

justinstoller (approved):

PR #31, opened 2024-05-22
(PE-38421) avoid stack overflow in event collection

jonathannewman:
See https://stuartsierra.com/2015/04/26/clojure-donts-concat

As was seen in the field, the event concatenation can cause a stack overflow for a large number of events.

This modifies the loop for retrieving events to use `into` instead of `concat` which prevents the stack overflow by collecting the results into a new vector with every iteration.


steveax (approved):
cat!
👍 LGTM!

PR #32, opened 2024-10-23
(PE-37592) Compress events for the same changed-path

justinstoller:
Previously, we would pass all events into a registered callback which triggered for the given watch-path. However, in some cases we may receive a `:create` event and then a `:modify` event as a new file is written to disk. We could also receive multiple `:modify` events, and because all of the events received from a WatchKey can be in any order we may receive the `:create` event after a `:modify` event. Eg, [`:modify`, `:create`, `:modify`, `:modify`] (imagine that these vectors are Events all sharing the same changed-path).

However, we want to compress modifies into any creation events. This will reduce the events given to callbacks. For example, the events [`:create`, `:modify`] will reduce to only [`:create`], while [`:modify`, `:create`] will also reduce to [`:create`]; events like [`:modify`, `:modify`, `:modify`] will reduce to [`:modify`]. We do not compress delete events, so events like [`:modify`, `:create`, `:modify`, `:delete`] will be reduced to [`:create`, `:delete`].

The tests had the pecular property of collecting events until a given event was received, then only returning _the given event_. The tests have been updated to now return all the events received. Which, without this patch, many tests fail because they receive modify and create events, when the test is only expecting a create event. The tests now pass with few changes because of the implemented event compression.

### DIFF
--- a/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
+++ b/src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
@@ -1,5 +1,6 @@
 (ns puppetlabs.trapperkeeper.services.watcher.filesystem-watch-core
-  (:require [clojure.tools.logging :as log]
+  (:require [clojure.set :as set]
+            [clojure.tools.logging :as log]
             [me.raynes.fs :as fs]
             [schema.core :as schema]
             [puppetlabs.i18n.core :refer [trs]]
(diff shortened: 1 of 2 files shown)


jonathannewman (approved):

src/clj/puppetlabs/trapperkeeper/services/watcher/filesystem_watch_core.clj
```diff
@@ -175,14 +176,60 @@
...
+          (update :tuples set/union #{(->tuple item)})
+          (update :events conj item))))
+
 (schema/defn process-events!
   "Process for side effects any events that occurred for watcher's watch-key"
   [watcher :- (schema/protocol Watcher)
    events :- [Event]]
-  (let [callbacks @(:callbacks watcher)]
+  (let [callbacks @(:callbacks watcher)
+        compressed-events (->> events (reduce reducible-remove {:tuples #{} :events []}) :events)]
```
donoghuc, line 229: Would all `events` have the same `changed-path`? In other words, is a collection of events here for a single file?

donoghuc: Answer is yes! We need to manage the compression per changed path.

donoghuc (approved):
I think the change to the tests is super valuable. It *seems* complex but I cant think of a simpler way to do it. I started thinking if grouping by `changed-path` first would simplify the set calculation but the transformations there probably are not any cleaner.

